PR #5, opened 2021-05-02
Tests to validate correlation calculations methods on pre-saved data samples.

maxhora:
common package is extended with common/time/time.go


levkoburburas (dismissed):

levkoburburas (approved):

Dicklesworthstone (approved):

common/time/time.go
```diff
@@ -0,0 +1,20 @@
+// Package time provides functions to work with execution time
+package time
```
levkoburburas, line 2: How about renaming this package to `runtime` ?

common/time/time.go
```diff
@@ -0,0 +1,20 @@
...
+
+// Measure retrieves name of the caller function and prints it along with elapsed time
+// Typical call of Measure should be deffered in the beggining of the function to measure
+func Measure(start time.Time) {
+	elapsed := time.Since(start)
+	pc, _, _, _ := runtime.Caller(1)
+	pcFunc := runtime.FuncForPC(pc)
+	funcNameOnly := regexp.MustCompile(`^.*\.(.*)$`)
+	funcName := funcNameOnly.ReplaceAllString(pcFunc.Name(), "$1")
+	fmt.Printf("\n%s took %s\n", funcName, elapsed)
```
levkoburburas, line 19: I think it would be more useful for all-purpose if the func returns string.

levkoburburas (commented):
Can you please fix it?

common/time/time.go
maxhora: @levkoburburas did you mean to return string instead of printing to std out?

so, the caller will invoke it like `defer fmt.Printf(runtime.Measure(time.now()))` ?

levkoburburas: Yes 
* Since this is common package, we shouldn't assume that all will use fmt.Printf, for example in my project I would use
`defer log.Debug(runtime.Measure(time.now())`
* And for more readable code, it is preferable to show that this command prints something. Otherwise `Measure` doesn't give me any clue what does it really do

maxhora: @levkoburburas got it, sounds great

levkoburburas (approved):
LGTM!

Dicklesworthstone (approved):

PR #10, opened 2021-05-03
X448 conn encryption

igorgark1593:
added x448 encryption and TransportCredentials


Dicklesworthstone (approved):

encryption/reader.go
```diff
@@ -0,0 +1,36 @@
...
+type StreamDecrypter struct {
+	Source io.Reader
+	Block  cipher.Block
+	Stream cipher.Stream
+}
+
+func (s *StreamDecrypter) Read(p []byte) (int, error) {
+	n, readErr := s.Source.Read(p)
+	if n > 0 {
+		s.Stream.XORKeyStream(p[:n], p[:n])
```
levkoburburas, line 18: https://golang.org/src/crypto/cipher/io.go?s=934:1010#L20
_A StreamWriter has no internal buffering;_

`cipher.NewCTR` uses 512 bytes for buffering, this is super ineffective for transferring large amounts of data

encryption/x448.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	return NewStreamEncrypter(c.shared[:aes.BlockSize], c.vi, writer)
+}
+
+// WrapReader methods wraps reader to decrypt stream
+func (c *X448Cipher) WrapReader(reader io.Reader) (io.Reader, error){
+	if len(c.vi) < aes.BlockSize{
+		return nil, fmt.Errorf("vi shouldn't be less than %d", aes.BlockSize)
+	}
+	return NewStreamDecrypter(c.shared[:aes.BlockSize], c.vi, reader)
```
levkoburburas, line 68: `aes.BlockSize` has size 16 bytes, that means AES-128. @ayk mentioned AES-512 [here](https://blockchain-devel.slack.com/archives/C0202VAM6Q3/p1619495800105800), AES-128 is be enough

levkoburburas (commented):
Let's summarize how this should be implemented: 
Handshake and data buffering (before encryption/decryption) must happen inside the net.Conn.
Google has already invented wheel, just use their [code](https://github.com/grpc/grpc-go/blob/6f35bbbfb82de348a1537774af2ffd706cd3bb12/credentials/alts/internal/conn/record.go#L146-L267) as example for writing your own.

Outside `net.Conn`, we need an implementation of data encryption/decryption, as is done [here](https://github.com/grpc/grpc-go/blob/6f35bbbfb82de348a1537774af2ffd706cd3bb12/credentials/alts/internal/conn/aes128gcm.go). Notice they use **_Seal_** 

> // Encrypt is the encryption function. dst can contain bytes at the beginning of
// the ciphertext that will not be encrypted but will be authenticated

PR #12, opened 2021-05-05
X448 conn encryption

igorgark1593:
moved inside net.Conn, added handshaker


Dicklesworthstone (approved):

encryption/conn.go
```diff
@@ -0,0 +1,217 @@
...
+	"net"
+)
+
+const (
+	// MsgLenFieldSize is the byte size of the frame length field of a
+	// framed message.
+	MsgLenFieldSize = 4
+	// The byte size of the message type field of a framed message.
+	msgTypeFieldSize = 4
+	// The bytes size limit for a ALTS record message.
```
levkoburburas, line 16: I don't think we have to mention `ALTS` in our code, because this means _(Application Layer Transport Security (ALTS) is a Google developed authentication and encryption system)_
* This is our implementation, yes we use their code, but we are using completely different logic.
* Our code is open source, people who is going to invest million dollars, (I believe) they will review thorough the code. Imagine how to surprise them, In other words, our code should look presentable.

encryption/conn.go
```diff
@@ -0,0 +1,217 @@
...
+// NewConn creates a new secure channel instance given the other party role and
+// handshaking result.
+func NewConn(c net.Conn, handshaker Handshaker) (net.Conn, error) {
+	var cipher Cipher
+	if handshaker.IsClient() {
+		pk, err := handshaker.ServerKeyExchange()
+		if err != nil {
+			return nil, err
+		}
+		cipher, err = handshaker.ClientKeyVerify(pk)
```
levkoburburas, line 58: You should pass `ClientKeyVerify` the Server public key that you received by negotiate, not client own. 
It seems that you do not understand how it should work.

encryption/handshaker_mock.go
```diff
@@ -0,0 +1,72 @@
+package encryption
```
levkoburburas, line 1: It looks like you just ignored my comments in the previous review.
If you use this mock just for testing, on your local environment, that's fine, but you can't commit this implementation.

encryption/proto/reverse.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
```
levkoburburas, line 1: the same, for what it should be in master branch?

levkoburburas (changes requested):
I stopped to review the all code, just no sense. It looks like you didn't read that I wrote in the previous. And did not understand the task

PR #21, opened 2021-05-08
test: add errors missing tests

al3rez:


common/errors/errors.go
```diff
@@ -30,6 +30,11 @@ func (err *Error) Unwrap() error {
 	return err.goerror.Err
 }
 
+// Goerr returns goerrors.Error
+func (err *Error) Goerr() *goerrors.Error {
```
levkoburburas, line 34: https://github.com/pastelnetwork/gonode/tree/master/common#errors

> Under the hood, the errors package is using the go-errors library, but this may change in the future, so the rest of the code should not depend on go-errors directly.

Unit test should test our package instead of `goerror`. Please don't refer to `goerror` at all.

common/errors/errors_test.go
```diff
@@ -0,0 +1,83 @@
+package errors_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/pastelnetwork/gonode/common/errors"
+)
+
+func TestNew(t *testing.T) {
```
levkoburburas, line 10: I see, you took it from `github.com/go-errors/errors/error_test.go` what's the point?

common/errors/errors_test.go
```diff
@@ -0,0 +1,83 @@
+package errors_test
+
+import (
```
levkoburburas, line 3: Please use `github.com/stretchr/testify`

levkoburburas (changes requested):
Thank you for the initiative, I appreciate it. Please fix issues.

common/errors/errors_test.go
al3rez: sure

common/errors/errors_test.go
al3rez: my idea is a code with test is better than one without. this can def. be improved this is mainly coverage not 100% functionality so more test cases can be added to expose errors package intention.

common/errors/errors.go
al3rez: @Levko errors package is  heavily tied to goerrors
mysuggestion would be to change the struct to accept error interface instead of goerrors.Error
that's why i needed to expose Goerr() to get underlying info

al3rez: I added Stack() to avoid exposing Goer(), so that we can change it to interface later.

levkoburburas: > @Levko errors package is heavily tied to goerrors
> mysuggestion would be to change the struct to accept error interface instead of goerrors.Error
> that's why i needed to expose Goerr() to get underlying info

You should use only those method which are used other packages, otherwise you test method that you wrote for unit_test, no sense. Why don't you use `Unwrap()` or compare string from `Error()`. Anyway just image that we don't use `goerrors`, and just pretend that you don't know what the Error struct contains in inside, pretend you know as much as the packages that use it.

> the struct to accept error interface instead of goerrors.Error
if you know better solution, please write your code, if it's really will be better than we have now, I will say you many thank you.

common/errors/errors_test.go
levkoburburas: I cannot agree with your opinion, because if have `_test.go` no one else will care about this unit test, all guys will think _this's already done_ and no one even thinks to improve it. I've known a lot of guys who like refactoring code, but I've never met anyone who would like to refactor a unit test.
So, I believe, if we write unit test it should really test our package, rather than imitation. If we don't have time for unit testing yet, it's best not to write anything. In this case, we will always remember this as what we must do. 
And of course, we should not throw words to the wind, plan but not do.

https://blockchain-devel.slack.com/archives/C01SLR99SAK/p1620239309035100?thread_ts=1620221825.025900&cid=C01SLR99SAK

common/errors/errors_test.go
levkoburburas: I see you have decided to do it your own way, by using `gotest.tools/assert`. you look like a rebel :) 

Ok Let's compare how a result looks with different tools:
"github.com/stretchr/testify/assert" (the package we decided you use for all unit tests)
```
--- FAIL: TestNew (0.00s)
    errors_test.go:15: 
        	Error Trace:	errors_test.go:15
        	Error:      	Not equal: 
        	            	expected: "foo"
        	            	actual  : "foofff"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-foo
        	            	+foofff
        	Test:       	TestNew
    errors_test.go:17: assertion failed: foo (string) != foofff (string)
FAIL
FAIL	github.com/pastelnetwork/gonode/common/errors	0.011s
FAIL
```

"gotest.tools/assert" (your choice)
```
--- FAIL: TestNew (0.00s)
    errors_test.go:16: assertion failed: foo (string) != foofff (string)
FAIL
FAIL	github.com/pastelnetwork/gonode/common/errors	0.010s
FAIL
```

Now please imagine, if we compare the large structure, how useful the detailed result can be for us.

al3rez: it's automatically imported using goimports, i didn't use choose it :)

levkoburburas (approved):
LGTM! Thanks!

Dicklesworthstone (approved):

Levko:
Hey levkoburburas you are a hacker and have violated my account.

Dicklesworthstone you are guilty of being a hacker.

Alireza Bashiri, Jeff Emmanuel your names are being passed to the authorities.

PR #27, opened 2021-05-11
Node connection improvement

levkoburburas:
* Load _Pastel ID_ from the `supernode` config file (instead of retrieving by pastel api)
* Improve `supernode` connections in the task.


a-ok123 (approved):

walletnode/services/artworkregister/task.go
```diff
@@ -84,7 +84,7 @@ func (task *Task) run(ctx context.Context) error {
 			case <-ctx.Done():
 				return errors.Errorf("task was canceled")
 			case <-node.conn.Done():
-				return errors.Errorf("prematurely closing the connection by %q", node.address)
+				return errors.Errorf("%q closed the connection", node.address)
```
a-ok123, line 87: may be "unexpectedly" ?

a-ok123 (changes requested):
Can we read pastel-api data from `pastel.conf`?  (somethings would be defaulted if missing from pastel.conf - like port)

Just a note: About node's own port in config - should be documented that port here must match the port in masternode.conf

walletnode/services/artworkregister/task.go
levkoburburas: Definitely, thanks for the hint, fixed.

levkoburburas:
> Can we read pastel-api data from `pastel.conf`? (somethings would be defaulted if missing from pastel.conf - like port)
> 
> Just a note: About node's own port in config - should be documented that port here must match the port in masternode.conf

Sure, agree, working on the implementation.

common/configurer/linux.go
```diff
@@ -0,0 +1,5 @@
+// +build linux
+
+package configurer
+
+var defaultConfigPaths = []string{".", "$HOME/.pastel"}
```
levkoburburas, line 5: Is it correct path for linux `"$HOME/.pastel"`?

common/configurer/windows.go
```diff
@@ -0,0 +1,5 @@
+// +build windows
+
+package configurer
+
+var defaultConfigPaths = []string{".", "$HOME/.pastel"}
```
levkoburburas, line 5: What is correct path for windows?

levkoburburas:
> Can we read pastel-api data from `pastel.conf`? (somethings would be defaulted if missing from pastel.conf - like port)
> 
> Just a note: About node's own port in config - should be documented that port here must match the port in masternode.conf

1. The highest priority of pastel-api settings is from walletnode config.
2. By default pastel (external) config `pastel.conf` is being looked for in the default path (depends on OS) and in the path of the walletnode binary.
3. The filepath to pastel (external) config can be also overwrited in CLI by flag `pastel-config-file` or in walletnode config:
```
pastel-api:
   config-file: /path/to/pastel.conf
```
4. If some settings  is missed in the walletnode config it takes from pastel (external) config.
```
pastel-api:
   port: 19932
```
the rest (hostname,username,password) will be taken from pastel (external) config.

5. If not specified neither in walletnode nor paste configs, it has default settings for (hostname and port).

levkoburburas:
You can also looking into [examples](https://github.com/pastelnetwork/gonode/tree/node-connection-improvement/walletnode/examples/configs)

levkoburburas:
One more, `node` has also default settings, so you can run without specified any configs, and it will works 
image

a-ok123 (approved):

PR #42, opened 2021-05-20
[CI] Cross-compile pqsignatures for Windows and MacOS.

maxhora:
xgo [forked](https://github.com/pastelnetwork/xgo) and tweaked to compile inside CIrcleCI docker executor;

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,44 +46,146 @@ commands:
           working_directory: ~/project/<< parameters.package >>
           command: |
             go test -v -race ./...
-
-  release-os:
+  build:
     parameters:
+      buildContainerName:
+        type: string
+      moduleSubDir:
+        type: string
       package:
         type: string
-      goos:
-        type: enum
-        enum: ["darwin", "linux", "windows"]
+      targets:
+        type: string
+      binaryName:
+        type: string
     steps:
+      - xgo-build:
+          sourcesContainerName: "sourcesContainer"
+          buildContainerName: << parameters.buildContainerName >>
+          moduleSubDir: << parameters.moduleSubDir >>
+          package: << parameters.package >>
+          targets: << parameters.targets >>
+          outputDirectory: "build"
       - run:
-          name: "<< parameters.package >>: Fetch dependencies"
-          working_directory: ~/project/<< parameters.package >>
+          name: "<< parameters.moduleSubDir >>: Validate existence of built binaries"
           command: |
-            go get -v ./...
-      - run:
-          name: "<< parameters.package >>: Build binary for << parameters.goos >>"
-          working_directory: ~/project/<< parameters.package >>
-          command: |
-            CGO_ENABLED=0 GOOS=<< parameters.goos >> GOARCH=amd64 go build -o bin/<< parameters.package >>-<< parameters.goos >>-amd64 -ldflags "-X github.com/pastelnetwork/gonode/common/version.version=$CIRCLE_TAG -extldflags '-static'" .
-      - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -113,6 +204,33 @@ jobs:
...
+      - setup_remote_docker:
+          version: 20.10.2
+          docker_layer_caching: true
+      - run:
+          name: Install xgo
+          command: |
+            go install -v github.com/pastelnetwork/xgo@48be1f35b934924c998997a8470ffdb6a022b968
+      - create-sources-container:
+          containerName: "sourcesContainer"
+      - build:
```
levkoburburas, line 216: Are you sure we need to build binaries in the test job? Because without it, checking all modules takes up to 10 minutes.

.circleci/config.yml
```diff
@@ -137,14 +255,37 @@ jobs:
...
+          buildContainerName: "walletnode"
+          moduleSubDir: "walletnode"
+          package: ""
+          targets: "linux/amd64,darwin-10.14/amd64,windows/amd64,"
+          binaryName: "walletnode"
+      - release:
+          buildContainerName: "supernode"
+          moduleSubDir: "supernode"
+          package: ""
+          targets: "linux/amd64,darwin-10.14/amd64,windows/amd64,"
```
levkoburburas, line 287: If we reduce list of the OS, does this cause an error in the `github-upload-all-os:` step ?

.circleci/config.yml
maxhora: I think it is better to wait than suddenly find that Windows or MacOS build is broken

.circleci/config.yml
maxhora: yes, `github-upload-all-os` will fail :)

.circleci/config.yml
levkoburburas: When we release apps, it will be built anyway. To be honest, even now it takes a lot of patience to wait until all modules are tested :) And when we add more tests, it will take her more time. Every time we commit, we run this flywheel. It eats our credits on paid plan

maxhora: @levkoburburas I believe it worth it; it just caught compilation error on Windows again.
New docker image pulling time is ~ 2m 20 seconds. Also 2m 30 seconds takes compilation of pqsignatures and walletnode for macos and windows; others projects and platforms are not cross-compiled at the moment.
I'm supposing that pqsignatures cross-compilation will be removed from ci once we will have legroast integrated into walletnode project.

Dicklesworthstone (approved):

.circleci/config.yml
maxhora: I've fixed similar issue for `build` command and its `Validate existence of built binaries`

levkoburburas: Just to clarify, if we change `targets` for supernode
```
targets: "linux/amd64,darwin-10.14/amd64,windows/amd64,"
```
to 
```
targets: "linux/amd64,"
```

in line 163 `github-release upload` will not return error ? when `github-upload-all-os` tries to upload a binary that doesn't exists?

.circleci/config.yml
levkoburburas: Sure Max, if you say it's worth, let's try to work with it.

.circleci/config.yml
maxhora: @levkoburburas my initial supposing was to use currently removed `release-os` command for linux-only builds.

but yes, uploading will fail if to do mentioned change.

levkoburburas: @maxhora Understood, if we need it in the future, we will probably have to transfer some part of the config to Makefile

levkoburburas (approved):
LGTM!

PR #44, opened 2021-05-20
common: errors package unit test #2

ashadi-cc:
- utils unit test 
- exitcode unit test


Dicklesworthstone (dismissed):

common/errors/utils_test.go
```diff
@@ -0,0 +1,106 @@
...
+		contain string
+	}{
+		{
+			err:     New("foo"),
+			contain: "New(\"foo\")",
+		}, {
+			err:     fmt.Errorf("bar"),
+			contain: "bar",
+		},
+	}
```
levkoburburas, line 25: please add one more case, to make sure `Unwrap` also tested. 
```go
fmt.Errorf("wrap another error: %w", errors.New("inner error")) 
```
BTW why I use `%w` instead of `%s` or  `%v`, https://blog.golang.org/go1.13-errors#TOC_3.3.

levkoburburas (changes requested):
One minor request and we can move on

common/errors/utils_test.go
ashadi-cc: ok got it. updated

levkoburburas (approved):
LGTM!

Dicklesworthstone (approved):

PR #47, opened 2021-05-24
Walletnode: artworkregister unit test

ashadi-cc:
- init artworkregister node unit test


levkoburburas (changes requested):
1. Create mocks files next to its interface.

e.g., if you create mock for `walletnode/node`
this mocks should be placed in `walletnode/node/mocks/...`

2. Don't need to push them into the repository. They should be generated while testing. For this you need add only `go generate` commands.

e.g., if we are talking about `walletnode/node/client.go`
```go
//go:generate mockery --name Client
//go:generate mockery --name Connection
//go:generate mockery --name RegisterArtwork

package node
```

when we need to run unit test, all we need to do s to run in the root `walletnode`
```
go generate ./...
```

Add in  `walletnode/.gitignore`
```
**/mocks
```

3. Create mocks for all interfaces that `Task` is using, including pastel.Client

walletnode/services/artworkregister/node_test.go
```diff
@@ -0,0 +1,400 @@
...
+			methodCall: methodCall{
+				connect:         1,
+				registerArtWork: 1,
+			},
+			mockArgs: mockArgs{mock.Anything, mock.AnythingOfType("string")},
+			args: args{
+				ctx: context.Background(),
+				err: nil,
+			},
+			assertion: assert.NoError,
```
levkoburburas, line 351: You use mock right, it's very good. 
But body of the test func very bloated and poorly readable.
Try to split into a separate functions. e.g. 

`newClient().AcceptAddress("127.0.0.1:4445").SomethingElse()...`
so these `&amp;mocks.Connection{}, &amp;mocks.RegisterArtwork{}` mocks should be inside, we don't need to shove them into cases explicitly, at least in this test.
In order to reuse this mock in other packages you can put into `/walletnode/node/test/client.go`

levkoburburas (commented):
You are doing everything right, you just need to simplify and make more readable. 
Create reusable mocks as a separate package to achieve thin tests.

walletnode/services/artworkregister/node_test.go
ashadi-cc: ok got it.

Dicklesworthstone (approved):

levkoburburas (approved):
Great work 👍

maxhora (approved):

PR #58, opened 2021-05-26
[ci] Release required jobs should have building on tags enabled; Publ…

maxhora:
…ish for specified targets; Improve build error on missed cross-compiled target executable.

- Repairs triggering of worflow on new GitHub release tag (each dependent job should have explicitly turned on running on new tags)
- Bash script command to upload only specified os targets on Release.
- supernode releasing limited only to build for Linux target.
- Improved error message in case if cross-compiled executable is not found.


Dicklesworthstone (approved):

levkoburburas (approved):
👍

PR #65, opened 2021-06-01
add config option for sqlite data source

al3rez:


p2p/kademlia/dht_test.go
```diff
@@ -366,9 +376,10 @@ func TestStoreAndFindLargeValue(t *testing.T) {
 
 	id1, _ := newID()
 	dht1, _ := NewDHT(ctx, getInMemoryStore(), &Options{
-		ID:   id1,
-		IP:   "127.0.0.1",
-		Port: 3000,
+		ID:             id1,
```
alonlong, line 379: i think u can try "github.com/stretchr/testify/suite" for the unit test, because i see many repeated hard code for many tests

p2p/config.go
```diff
@@ -17,6 +17,9 @@ type Config struct {
 	BootstrapPort string `mapstructure:"bootstrap_port" json:"bootstrap_port,omitempty"`
 	// Use STUN protocol for public addr discovery
 	UseStun bool `mapstructure:"use_stun" json:"use_stun,omitempty"`
+	// a driver-specific data source name, usually consisting of
+	// at least a database name and connection information.
+	DataSourceName string `mapstructure:"data_source_name" json:"data_source_name,omitempty"`
```
levkoburburas, line 22: Hint: `supernode` must also run without configuration.

levkoburburas (changes requested):
Please fix minor config issue

p2p/config.go
levkoburburas: I guess `DataSourceName` should be called `DataDir`.

`DataDir` by default should has the path  `work-dir`/p2p, e.g. [example](https://github.com/pastelnetwork/gonode/blob/fcf89eae0d2a649caee1adbd6384940513d4b2b8/supernode/cmd/app.go#L129). Use `ioutil.TempDir` for unit tests

p2p/p2p.go
```diff
@@ -110,7 +111,8 @@ func (service *p2p) configure(ctx context.Context) error {
 }
 
 // New returns a new p2p instance.
-func New(config *Config) P2P {
+func New(config *Config, dataDir string) P2P {
+	config.DataDir = dataDir
```
levkoburburas, line 115: more refined way

```go
// SetDefaultDataDir assigns the value to DataDir if it wasn't specified in the config file.
func (config *Config) SetDefaultDataDir(dir string) {
	if config.DataDir == "" {
		config.DataDir = dir
	}
}
```

p2p/kademlia/dht.go
```diff
@@ -89,7 +93,7 @@ func NewDHT(ctx context.Context, store dao.Key, options *Options) (*DHT, error)
 	dht.ht = ht
 	dht.networking = &realNetworking{}
 
-	if err = store.Init(ctx); err != nil {
+	if err = store.Init(ctx, options.DataDir); err != nil {
```
levkoburburas, line 96: I think sqlite needs to get a file, not a directory.
So, I think you need to add something like

```go
const dbFilename = "p2p.sqlite"
......
filepath.Join(options.DataDir, dbFilename)
```

supernode/examples/configs/localnet-4444.yml
```diff
@@ -13,5 +13,6 @@ node:
 p2p:
   listen_addresses: "0.0.0.0"
   port: 4454
+  data_source_name: ":memory:"
```
levkoburburas, line 16:
```suggestion
  # default value is assigned automatically, by concatenated flag `--work-dir` and `p2p` as a subfolder
  data_dir: "path/to/p2p/data/dir"

  # if not specified, persistent database store is used
  memory_db: true

 ```

levkoburburas (approved):
I had to change config files, but everything else is done well. Thanks.

Dicklesworthstone (approved):

supernode/examples/configs/localnet-4444.yml
al3rez: sorry I misunderstand here for config files :) thanks for fixing them.

PR #67, opened 2021-06-03
Create metadb service for supernode

alonlong:


metadb/rqlite/cluster/join.go
```diff
@@ -43,14 +44,14 @@ func Join(srcIP string, joinAddr []string, id, addr string, voter bool, numAttem
 				return j, nil
 			}
 		}
-		logger.Printf("failed to join cluster at %s: %s, sleeping %s before retry", joinAddr, err.Error(), attemptInterval)
+		logger.Errorf("failed to join cluster at %s: %s, sleeping %s before retry", joinAddr, err.Error(), attemptInterval)
 		time.Sleep(attemptInterval)
 	}
-	logger.Printf("failed to join cluster at %s, after %d attempts", joinAddr, numAttempts)
+	logger.Errorf("failed to join cluster at %s, after %d attempts", joinAddr, numAttempts)
```
levkoburburas, line 50:
```suggestion
	logger.Warnf("failed to join cluster at %s, after %d attempts", joinAddr, numAttempts)
```

elsewhere it might be worth changing as well

metadb/rqlite/disco/client.go
```diff
@@ -7,9 +7,10 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"net/http"
-	"os"
+
```
levkoburburas, line 11:
```suggestion
	"github.com/pastelnetwork/gonode/common/errors"
```
Again, Since we have to maintain this code and catch bugs and errors, it is preferable to have a stack error of the possible errors.

metadb/rqlite/tcp/mux_test.go
```diff
@@ -24,9 +24,9 @@ func TestMux(t *testing.T) {
 	if err := quick.Check(func(n uint8, msg []byte) bool {
 		if testing.Verbose() {
 			if len(msg) == 0 {
-				log.Printf("n=%d, <no message>", n)
+				log.Infof("n=%d, <no message>", n)
```
levkoburburas, line 27:
```suggestion
				t.Logf("n=%d, <no message>", n)
```

metadb/rqlite/tcp/mux_test.go
```diff
@@ -24,9 +24,9 @@ func TestMux(t *testing.T) {
...
 			} else {
-				log.Printf("n=%d, hdr=%d, len=%d", n, msg[0], len(msg))
+				log.Infof("n=%d, hdr=%d, len=%d", n, msg[0], len(msg))
```
levkoburburas, line 29:
```suggestion
				t.Logf("n=%d, hdr=%d, len=%d", n, msg[0], len(msg))
```

metadb/rqlite/tcp/mux_test.go
```diff
@@ -41,7 +41,8 @@ func TestMux(t *testing.T) {
 		}
 		mux.Timeout = 200 * time.Millisecond
 		if !testing.Verbose() {
-			mux.Logger = log.New(ioutil.Discard, "", 0)
+			mux.Logger = log.DefaultLogger.WithField("prefix", "mux_test")
```
levkoburburas, line 44: It's really weird to use standard log in unit test. I understand, it's not your idea. but it is better to implement it in a more correct way, using `t.Logf` If we have to change the code anyway

metadb/service.go
```diff
@@ -0,0 +1,31 @@
+package metadb
+
+import (
+	"context"
+
+	"github.com/pastelnetwork/gonode/metadb/rqlite/store"
+)
+
+const (
+	logPrefix = "rqlite"
```
levkoburburas, line 10:
```suggestion
	logPrefix = "metadb"
```

metadb/server.go
```diff
@@ -0,0 +1,337 @@
...
+		log.WithContext(ctx).Infof("join addresses are: %v", joins)
+		advAddr := s.config.RaftAddress
+		if s.config.RaftAdvertiseAddress != "" {
+			advAddr = s.config.RaftAdvertiseAddress
+		}
+
+		// try to parse the join duration
+		joinDuration, err := time.ParseDuration(s.config.JoinInterval)
+		if err != nil {
+			return errors.Errorf("parse JoinInterval: %v", err)
```
levkoburburas, line 277:
```suggestion
			return errors.Errorf("parse JoinInterval: %w", err)
```
always use `%w` with errors, to not lose already the wrapped errors

metadb/server.go
```diff
@@ -0,0 +1,337 @@
...
+
+	// open the cluster service
+	if err := c.Open(); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+// start the rqlite server, and try to join rqlite cluster if the join addresses is not empty
+func (s *Service) startServer(ctx context.Context) error {
```
levkoburburas, line 179: too long function, better to split it by meaning

metadb/server.go
```diff
@@ -0,0 +1,337 @@
...
+		}
+
+		// try to parse the join duration
+		joinDuration, err := time.ParseDuration(s.config.JoinInterval)
+		if err != nil {
+			return errors.Errorf("parse JoinInterval: %v", err)
+		}
+
+		tlsConfig := tls.Config{InsecureSkipVerify: s.config.NoVerify}
+		if s.config.X509CACert != "" {
```
levkoburburas, line 281: Do we need this in our project? please ask @ayk

metadb/store.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/pastelnetwork/gonode/common/errors"
+	"github.com/pastelnetwork/gonode/metadb/rqlite/command"
+)
+
+const (
+	// ReadLevelNono - with none, the node simply queries its local SQLite database, and does not even check if it is the leader
+	ReadLevelNono = "nono"
+	// ReadLevelWeek - weak instructs the node to check that it is the leader, before querying the local SQLite file
+	ReadLevelWeek = "week"
```
levkoburburas, line 15:
```suggestion
	// ReadLevelWeak - weak instructs the node to check that it is the leader, before querying the local SQLite file
	ReadLevelWeak = "weak"
```

metadb/store_test.go
```diff
@@ -0,0 +1,93 @@
...
+	if s.cancel != nil {
+		s.cancel()
+	}
+	s.wg.Wait()
+
+	// remove the data directy
+	os.RemoveAll(s.config.DataDir)
+}
+
+func (s *testSuite) TestWriteAndQuery() {
```
levkoburburas, line 59: We use [table driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) in all our unit tests 
please also use this package https://github.com/stretchr/testify
Examine other our unit tests to understand better.

supernode/configs/node.go
```diff
@@ -10,12 +11,14 @@ type Node struct {
 	// `squash` field cannot be pointer
 	ArtworkRegister artworkregister.Config `mapstructure:",squash" json:"artwork_register,omitempty"`
 	Server          *server.Config         `mapstructure:"server" json:"server,omitempty"`
+	MetaDB          *metadb.Config         `mapstructure:"metadb" json:"metadb,omitempty"`
```
levkoburburas, line 14: I think it's better to move it to a higher level in the `Config` structure

metadb/service.go
```diff
@@ -0,0 +1,31 @@
...
+	ready  chan struct{} // mark the rqlite node is started
+}
+
+// NewService returns a new service for rqlite cluster
+func NewService(config *Config) *Service {
+	return &Service{
+		config: config,
+		ready:  make(chan struct{}, 1),
+	}
+}
```
levkoburburas, line 26:
```suggestion
// Service represents the rqlite cluster
type Service interface {
	....
}

type service struct {
	config *Config       // the service configuration
	db     *store.Store  // the store for accessing the rqlite cluster
	ready  chan struct{} // mark the rqlite node is started
}

// NewService returns a new service for rqlite cluster
func New(config *Config) Service {
	return &service{
		config: config,
		ready:  make(chan struct{}, 1),
	}
}
```

When we write unit test for business logic services, we use mock, rather than really DB

levkoburburas (changes requested):
Overall quality work, but need to meet to our requirements. I mentioned in the comments.

metadb/config.go
```diff
@@ -0,0 +1,132 @@
...
+	defaultRaftLeaderLeaseTimeout = "0s"
+	defaultRaftShutdownOnRemove   = false
+	defaultRaftLogLevel           = "error"
+	defaultCompressionSize        = 150
+	defaultCompressionBatch       = 5
+	defaultOnDisk                 = false
+	defaultDataDir                = ".data"
+	defaultNodeX509Cert           = "cert.pem"
+	defaultNodeX509Key            = "key.pem"
+)
```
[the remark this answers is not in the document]
alonlong: now these fileds can be configured in yaml file by an user, Agree with u only have settings that the user can actually change. I think we can control it by the Config structure, like

metadb/rqlite/cluster/service.go
```diff
@@ -68,14 +69,14 @@ func New(tn Transport) *Service {
 		tn:      tn,
 		addr:    tn.Addr(),
 		timeout: 10 * time.Second,
-		logger:  log.New(os.Stderr, "[cluster] ", log.LstdFlags),
+		logger:  log.DefaultLogger.WithField("prefix", "cluster"),
```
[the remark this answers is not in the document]
alonlong: done

metadb/rqlite/cluster/service.go
```diff
@@ -68,14 +69,14 @@ func New(tn Transport) *Service {
...
 	}
 }
 
 // Open opens the Service.
 func (s *Service) Open() error {
 	go s.serve()
-	s.logger.Println("service listening on", s.tn.Addr())
+	s.logger.Infof("service listening on: %v", s.tn.Addr())
```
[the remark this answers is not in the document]
alonlong: done

metadb/rqlite/cluster/join.go
```diff
@@ -43,14 +44,14 @@ func Join(srcIP string, joinAddr []string, id, addr string, voter bool, numAttem
 				return j, nil
 			}
 		}
-		logger.Printf("failed to join cluster at %s: %s, sleeping %s before retry", joinAddr, err.Error(), attemptInterval)
+		logger.Errorf("failed to join cluster at %s: %s, sleeping %s before retry", joinAddr, err.Error(), attemptInterval)
```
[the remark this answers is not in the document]
alonlong: done

metadb/rqlite/cluster/join.go
alonlong: i think it's better to keep error here, because after attempts, it will cause the application exit when starting the supernode

metadb/rqlite/disco/client.go
alonlong: done

metadb/rqlite/tcp/mux_test.go
alonlong: done

metadb/rqlite/tcp/mux_test.go
alonlong: done

metadb/rqlite/tcp/mux_test.go
alonlong: done

metadb/service.go
alonlong: done

metadb/server.go
alonlong: done

metadb/config.go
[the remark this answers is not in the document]
levkoburburas: Agree

supernode/configs/node.go
alonlong: Agree

metadb/server.go
alonlong: done

metadb/store.go
alonlong: done

supernode/configs/node.go
alonlong: done

metadb/service.go
alonlong: done

metadb/rqlite/cluster/service.go
```diff
@@ -59,7 +60,7 @@ type Service struct {
 	https   bool   // Serving HTTPS?
 	apiAddr string // host:port this node serves the HTTP API.
 
-	logger *log.Logger
+	logger *logrus.Entry
```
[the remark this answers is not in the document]
alonlong: done

metadb/store_test.go
alonlong: done

metadb/server.go
alonlong: done

Dicklesworthstone (approved):

levkoburburas (changes requested):
Good job 👍  
There are minor issues left to fix and we can go with it.

But first, let's merge [this](https://github.com/pastelnetwork/gonode/pull/75/files)
1. We cannot allow someone else's logger to print to our console. There are a lot of reasons, formatting, writing in log file etc.
2.
```go
import "github.com/pastelnetwork/gonode/common/errors"

var ErrNotLeader = errors.New("not leader")

// Execute executes queries that return no rows, but do modify the database.
func (s *Store) Execute(ex *command.ExecuteRequest) ([]*sql.Result, error) {
	if s.raft.State() != raft.Leader {
		return nil, ErrNotLeader
	}
	return s.execute(ex)
}
``` 
It ruins stack error, and it will contains stack error line where you are creating the custom error, rather than where you are really returning it

levkoburburas:
One more about log level. In many case you use `Info` instead of `Debug`. 
Better to use `Debug` by default for everything you are in doubt, and then during testing, if you understand that it should be `Info level`, rename.

metadb/store_test.go
```diff
@@ -0,0 +1,70 @@
...
+	// insert the value to table
+	insertion = "insert into foo(id, name) values(2,'belayu')"
+	result, err = s.Write(ctx, insertion)
+	assert.Nil(t, err, "execute insertion statement")
+	assert.Empty(t, result.Error, "result after execution")
+
+	// select a row from table with week level
+	selection := "select * from foo where id = 1"
+	rows, err := s.Query(ctx, selection, "weak")
+	assert.Nil(t, err, "query statements")
```
levkoburburas, line 54: It should be written with table driven tests [article](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) and [example](https://github.com/pastelnetwork/gonode/blob/master/common/errors/error_test.go)

metadb/service.go
```diff
@@ -0,0 +1,41 @@
...
+// Service represents the rqlite cluster
+type Service interface {
+	// Run starts the rqlite service
+	Run(ctx context.Context) error
+	// Query execute a query, not support multple statements
+	// level can be 'none', 'weak', and 'strong'
+	Query(ctx context.Context, statement string, level string) (*QueryResult, error)
+	// Write execute a statement, not support multple statements
+	Write(ctx context.Context, statement string) (*WriteResult, error)
+}
```
levkoburburas, line 22: Oh, I have should explained it in the previous reviews.

The `metadb` package must provide DAO (data access object) instead of sql query executor. 
Business logic services do not need to know anything about how and where data is stored, and even more so, they do not need to compose sql queries. If tomorrow we change the type of db, we should not touch the code of the business services.

It should look something like this:
```go
StoreArtworksData(ctx context.Context,  data *SomeDataStruct) error
GetArtworksData(ctx context.Context,  txid string) (*SomeDataStruct, error)
```
But it's just example, I don't know exactly what we are going to store, in details. To clarify this, ask Jeff or aik

levkoburburas (changes requested):
What I noticed. 
`rqlite` code we have almost integrated, but `metadb` service we didn't implemented yet

metadb/config.go
```diff
@@ -0,0 +1,38 @@
+package metadb
+
+const (
+	defaultHTTPAddress  = "0.0.0.0:4001"
+	defaultRaftAddress  = "0.0.0.0:4002"
+	defaultRaftLogLevel = "error"
```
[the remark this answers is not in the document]
alonlong: done

metadb/service.go
alonlong: yeah, it should be the next task, i will ask @jeff and @ayk

metadb/store_test.go
alonlong: done

metadb/service.go
Dicklesworthstone: Yes please do it, you can create the task yourself in Freedcamp.

metadb/store_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"context"
+	"fmt"
+	"os"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWriteAndQuery(t *testing.T) {
```
levkoburburas, line 13: Please split it into two Tests
```go
TestWrite(
TestQuery(
```

levkoburburas (commented):
1. Please add default config settings into `examples/configs/....` files
2. https://github.com/pastelnetwork/gonode/pull/79

metadb/service.go
alonlong: @Dicklesworthstone already create a task

metadb/store_test.go
alonlong: ok, Good idea

alonlong: done

metadb/config.go
```diff
@@ -0,0 +1,34 @@
...
+	RaftAddress string `mapstructure:"raft_address" json:"raft_address,omitempty"`
+	// comma-delimited list of nodes, through which a cluster can be joined (proto://host:port)"
+	JoinAddress string `mapstructure:"join_address" json:"join_address,omitempty"`
+	// set Discovery Service URL
+	DiscoveryURL string `mapstructure:"discovery_url" json:"discovery_url,omitempty"`
+	// set Discovery ID. If not set, Discovery Service not used
+	DiscoveryID string `mapstructure:"discovery_id" json:"discovery_id,omitempty"`
+	// data directory for rqlite
+	DataDir string `mapstructure:"data_dir" json:"data_dir,omitempty"`
+}
```
levkoburburas, line 25:
```suggestion
type Config struct {
         // IPv4 or IPv6 address for http server and rafd
	ListenAddress string `mapstructure:"listen_address" json:"listen_address,omitempty"`
	
	// http server bind port. For HTTPS, set X.509 cert and key
	HTTPPort string `mapstructure:"http_port" json:"http_port,omitempty"`
	
	// raft port communication bind address
	RaftPort string `mapstructure:"raft_port" json:"raft_port,omitempty"`	
		
	// unique name for node. If not set, set to hostname
	NodeID string `mapstructure:"node_id" json:"node_id,omitempty"`
	
	// set Discovery Service URL
	DiscoveryURL string `mapstructure:"discovery_url" json:"discovery_url,omitempty"`
	
	// set Discovery ID. If not set, Discovery Service not used
	DiscoveryID string `mapstructure:"discovery_id" json:"discovery_id,omitempty"`
	
	// data directory for rqlite
	DataDir string `mapstructure:"data_dir" json:"data_dir,omitempty"`
}
```

1. Supernode is assumed to work in the same IP address, it makes no sense to give the opportunity to specify different HTTP and Raft addresses. Moreover I think we should to reduce settings to give the opportunity to specify only one listen_address for all services. I will do it later. but you please leave only one `ListenAddress`. Don't forget to update configs.

2. Not need to specify `JoinAddress`. There is no centralized server, you need to get the list of supernodes from Pastel RPC API, and try to connect automatically. For now please remove it. But create a task in Freedcamp, `Implement connection to rqlite cluster`

alonlong: task is created

alonlong: done

alonlong (commented):
done

metadb/server.go
```diff
@@ -0,0 +1,233 @@
...
+	// open the cluster service
+	if err := c.Open(); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+// create and open the store of rqlite cluster
+func (s *service) initStore(ctx context.Context, raftTn *tcp.Layer) (*store.Store, error) {
+	raftDir := filepath.Join(s.workDir, s.config.DataDir)
```
levkoburburas, line 96: I meant different. 
By default it should be like this

take an example from 
https://github.com/pastelnetwork/gonode/blob/add54140e3bcb36806bad451d14c37a115cf920c/p2p/config.go#L27-L32
https://github.com/pastelnetwork/gonode/blob/add54140e3bcb36806bad451d14c37a115cf920c/supernode/cmd/app.go#L133

but an user can specify another dir, anywhere in file system. 
for example, to place the database to a raid disk, or whatever.

levkoburburas (approved):

maxhora (approved):

PR #69, opened 2021-06-03
receive & process GET artwork search request

mateeullahmalik:
- Design art search get request & "update" it to websocket connection
- Task: https://freedcamp.com/view/2956460/tasks/40066506
- Search for art activation tickets in blockchhain
- https://freedcamp.com/view/2956460/tasks/40066495


walletnode/api/design/artworks.go
```diff
@@ -307,3 +340,190 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+	})
+	Attribute("limit", Int, func() {
+		Description("Number of results to be return")
+		Minimum(10)
+		Default(10)
+		Example(10)
+	})
+
+	Attribute("artist_name", String, func() {
+		Description("search for matches in ticket's field - artist_name")
```
levkoburburas, line 358:
```suggestion
		Description("Name of the artist")
```
The aim of the Description is to describe the field. This is not endpoint. Please fix it everywhere

walletnode/api/design/artworks.go
```diff
@@ -307,3 +340,190 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+	Attribute("keyword", String, func() {
+		Description("search for matches in ticket's field - artwork_keyword_set")
+		MaxLength(256)
+	})
+
+	Attribute("min_block", Int, func() {
+		Description("return art with art min_block <= activation ticket txid ")
+	})
+
+	Attribute("max_block", Int, func() {
```
levkoburburas, line 383: Where validation? Will it return an error if I ask `max_block=-1111`?  please fix it everywhere

walletnode/api/design/artworks.go
```diff
@@ -307,3 +340,190 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
 	})
 	Required("taskId")
 })
+
+// SearchArtworkParams are query params to searchArtwork request
+var SearchArtworkParams = func() {
```
levkoburburas, line 345: Required fields must be listed
```go
Required("artist_name", ..........)
```

walletnode/api/design/artworks.go
```diff
@@ -307,3 +340,190 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+		Description("return art with min_rareness_score <= rareness_score ")
+	})
+
+	Attribute("max_rareness_score", Int, func() {
+		Description("return art wit rareness_score <= max_rareness_score")
+	})
+}
+
+// ArtworkSearchResult  is response of artwork search.
+var ArtworkSearchResult = Type("ArtworkSearchResult", func() {
```
levkoburburas, line 413: Unfortunately you didn't understand what need to be returned, re-read @ayk [message]( https://blockchain-devel.slack.com/archives/C022CA67U69/p1622173995106800) and ask him if it's not clear for you

levkoburburas (changes requested):
Need to fix the design

walletnode/api/design/artworks.go
mateeullahmalik: but these are filter query params , still?

mateeullahmalik:
Thanks for your suggestions @levkoburburas - I've resolved them & added next commit on it which refers to next sub-task of artSearch ticket.
 https://freedcamp.com/view/2956460/tasks/40066495
 pls provide your feedback

walletnode/services/artworksearch/task.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// TODO: iterate through tickets & get reg tickets
+	// In parallel; filter reg tickets based on search params
+	// & fetch thumbnail as soon as one matches
+	// send the thumbnail along with data through resultChan
+
+	return nil
+}
+
+// SubscribeSearchResult returns a new search resultof the state.
```
mateeullahmalik, line 56: I need to update this comment.

pastel/art_ticket.go
```diff
@@ -1,7 +1,31 @@
 package pastel
 
-// ArtTicket represents pastel art ticket.
+// Ticket is Pastel NFT ticket structure
+type Ticket struct {
```
levkoburburas, line 4: You cannot occupy name `Ticket` because Pastel has different tickets. for example `IDTicket`.
All structures related to Art must be prefixed with `Art`.

```suggestion
type ArtTicket struct {
     ArtTicketData ArtTicketData `json:"-"`
}

type ArtTicketData struct {
     AppTicketData ArtAppTicket `json:"-"`
}
```

pastel/act_ticket.go
```diff
@@ -0,0 +1,24 @@
...
+
+// ActivationTicketType represents a type of the activation ticket.
+type ActivationTicketType string
+
+// List of types of activation ticket.
+const (
+	ActTicketAll       ActivationTicketType = "all"
+	ActTicketAvailable ActivationTicketType = "available"
+	ActTicketSold      ActivationTicketType = "sold"
+)
```
levkoburburas, line 11: be consistent, use `Act` or  `Activation` everywhere.
I personally prefer `Act` to reflect API method

pastel/client.go
```diff
@@ -102,6 +102,28 @@ func (client *client) Verify(ctx context.Context, data []byte, signature, pastel
...
+
+	if err := client.callFor(ctx, &tickets, "tickets", "list", "act", actType, minHeight); err != nil {
+		return nil, errors.Errorf("failed to get id tickets: %w", err)
+	}
+
+	return tickets, nil
+}
+
+// GetRegTicket implements pastel.Client.GetRegTicket
+func (client *client) GetRegTicket(ctx context.Context, regTxid string) (*Ticket, error) {
```
levkoburburas, line 117: We omit `Get` in Golang 

```suggestion
func (client *client) RegTicket(ctx context.Context, regTxid string) (*Ticket, error) {
```

walletnode/api/design/artworks.go
```diff
@@ -119,6 +119,45 @@ var _ = Service("artworks", func() {
...
+	Method("artSearch", func() {
+		Description("Streams the search result for artwork")
+		Meta("swagger:summary", "Streams the search result for Artwork")
+
+		Payload(SearchArtworkParams)
+		StreamingResult(func() {
+			Extend(ArtworkTicket)
+			Attribute("image", Bytes, func() {
+				Description("Thumbnail image")
+			})
```
levkoburburas, line 132: Why not move `image` into `ArtworkTicket` type?

walletnode/api/design/artworks.go
```diff
@@ -307,3 +346,93 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+		MaxLength(256)
+	})
+
+	Attribute("keyword", String, func() {
+		Description("keyword is one of the values in artwork_keyword_set")
+		MaxLength(256)
+	})
+
+	Attribute("min_block", Int, func() {
+		Description("Minimum blocknum ")
```
levkoburburas, line 387:
```suggestion
		Description("Minimum blocknum")
```

walletnode/api/design/artworks.go
```diff
@@ -307,3 +346,93 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+		Description("Name of the artist")
+		MaxLength(256)
+		Example("Mona Lisa")
+	})
+	Attribute("art", String, func() {
+		Description("Art is artwork title")
+		MaxLength(256)
+	})
+	Attribute("series", String, func() {
+		Description("series refers to artwork series name")
```
levkoburburas, line 373:
```suggestion
		Description("Artwork series name")
```

walletnode/api/design/artworks.go
```diff
@@ -307,3 +346,93 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+		MaxLength(256)
+	})
+	Attribute("descr", String, func() {
+		Description("descr refers to artist written statement")
```
levkoburburas, line 377:
```suggestion
		Description("Artist written statement")
```

walletnode/api/design/artworks.go
```diff
@@ -307,3 +346,93 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+		MaxLength(256)
+	})
+
+	Attribute("keyword", String, func() {
+		Description("keyword is one of the values in artwork_keyword_set")
```
levkoburburas, line 382: I believe you can write better Description, and with Capitalize first letter

walletnode/api/design/artworks.go
```diff
@@ -307,3 +346,93 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+	})
+
+	Attribute("min_block", Int, func() {
+		Description("Minimum blocknum ")
+		Minimum(1)
+		Default(1)
+	})
+
+	Attribute("max_block", Int, func() {
+		Description("maximum blocknum")
```
levkoburburas, line 393:
```suggestion
		Description("Maximum blocknum")
```

walletnode/api/design/artworks.go
```diff
@@ -307,3 +346,93 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+
+	Attribute("min_copies", Int, func() {
+		Description("Minimum number of created copies")
+		Minimum(1)
+		Maximum(1000)
+		Example(1)
+	})
+
+	Attribute("max_copies", Int, func() {
+		Description("Maximum number of crated copies")
```
levkoburburas, line 405:
```suggestion
		Description("Maximum number of created copies")
```

walletnode/api/design/artworks.go
```diff
@@ -307,3 +346,93 @@ var RegisterTaskPayload = Type("RegisterTaskPayload", func() {
...
+
+	Attribute("max_nsfw_score", Int, func() {
+		Description("Maximum nsfw score")
+		Minimum(1)
+		Maximum(1000)
+		Example(1000)
+	})
+
+	Attribute("min_rareness_score", Int, func() {
+		Description("Minimum rareness score ")
```
levkoburburas, line 426:
```suggestion
		Description("Minimum rareness score")
```

levkoburburas (changes requested):
Please reply to every comment. By showing that you took it into account or If you do not agree, please give your point of view.

walletnode/services/artworksearch/service.go
```diff
@@ -0,0 +1,64 @@
...
+	return tasks
+}
+
+// Task returns the task of the artwork search by the given id.
+func (service *Service) Task(id string) *Task {
+	return service.Worker.Task(id).(*Task)
+}
+
+// AddTask runs a new task of the artwork search and returns its taskID.
+func (service *Service) AddTask(request *artworks.ArtSearchPayload) string {
```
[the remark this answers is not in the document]
mateeullahmalik: @levkoburburas  I had this at the back of my mind but was a bit confused about how we take duplication at the expense of "good architecture". I'll change this, thanks for the suggestion.

walletnode/services/artworksearch/status.go
```diff
@@ -0,0 +1,38 @@
...
+func (status Status) String() string {
+	if name, ok := statusNames[status]; ok {
+		return name
+	}
+	return ""
+}
+
+// IsFinal returns true if the status is the final.
+func (status Status) IsFinal() bool {
+	return status == StatusTaskCompleted
```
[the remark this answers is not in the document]
mateeullahmalik: Alright.

walletnode/services/artworksearch/task.go
```diff
@@ -0,0 +1,89 @@
...
+	return task.resultChan
+}
+
+func (task *Task) actTickets(ctx context.Context) (tickets pastel.ActivationTickets, err error) {
+	actTickets, err := task.pastelClient.ActTickets(ctx, pastel.ActTicketAll, task.Request.MinBlock)
+	if err != nil {
+		return tickets, err
+	}
+
+	for i := 0; i < len(actTickets); i++ {
```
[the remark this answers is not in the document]
mateeullahmalik: range creates a copy of entire list which I try to avoid. range is readable but at the expense of time & memory
But I'll make this change

walletnode/services/artworksearch/status.go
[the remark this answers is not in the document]
mateeullahmalik: I agree there is some infra to be built
 This part is WIP.

walletnode/api/design/artworks.go
mateeullahmalik: This is a design decision depending on what user experience we want to provide on Wallet UI. I've asked several times. Can you help with this? I personally don't see any field that "should" be required. 
for minimum block, I put a default value.
@levkoburburas

walletnode/api/design/artworks.go
mateeullahmalik: @levkoburburas I was trying to write in terms of what query param means..as opposed to what the related field means in a domain entity ..as that is something that has already been written.
but I'll try to improve this.

walletnode/api/services/artwork.go
```diff
@@ -149,11 +151,36 @@ func (service *Artwork) Mount(ctx context.Context, mux goahttp.Muxer) goahttp.Se
...
+			res := &artworks.ArtSearchResult{
+				Name: ticket.ArtTicketData.AppTicketData.ArtworkTitle,
+				// TODO: Add rest of the fields
+			}
+			if err := stream.Send(res); err != nil {
+				return artworks.MakeInternalServerError(err)
+			}
+
+		}
+	}
```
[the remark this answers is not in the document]
mateeullahmalik: @levkoburburas I agree there is some infra to be built.  At this point I was trying to be as simple as possible. (just try to focus on what is required in given sub-task) Thanks for sharing this. will make the changes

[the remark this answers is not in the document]
mateeullahmalik: @levkoburburas I agree there is some infra to be built.  At this point I was trying to be as simple as possible. (just try to focus on what is required in given sub-task) Thanks for sharing this. will make the changes

walletnode/api/design/artworks.go
levkoburburas: Without a required field, the search becomes useless. You must indicate what you are looking for.
I don't know, it isn't my scope. Try to be more persistent with question to @ayk @jeff. There is no other way.

walletnode/services/artworksearch/task.go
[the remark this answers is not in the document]
levkoburburas: I know, but it depends on the situation, actually I don't know how much data are you going to process.
But if we judge that you are going to receive them through the API, for sure that not hundreds of thousands, otherwise we should add a filter to the API itself

pastel/art_ticket.go
mateeullahmalik: image
This is the outer ticket structure which has a field art_ticket. If I name the outer ticket as art_ticket then what will we do with the inner one? Should we name the outer as PastelTicket or rename the inner as ArtRegTicket?

walletnode/api/design/artworks.go
mateeullahmalik: @levkoburburas I'm using an already existing object in this file, I'm reusing that & with this extend ...the image essentially becomes a part of the object for this API

pastel/art_ticket.go
levkoburburas: I advice you to work with real data. from `pastel-cli`, because in fact it can turn out a little differently. Or how are you going to test it? 
So, please show the result of Pastel API, and then we can continue to discuss.

levkoburburas: So far you can use `json` tag according to documented structure, but the Golang structure should be called correctly 

example:
```go
type ArtTicket struct {
     ......
     ArtTicketData ArtTicketData `json:"art_ticket"`
}
```

pastel/client.go
```diff
@@ -102,6 +102,28 @@ func (client *client) Verify(ctx context.Context, data []byte, signature, pastel
...
+	if err := client.callFor(ctx, &tickets, "tickets", "list", "act", actType, minHeight); err != nil {
+		return nil, errors.Errorf("failed to get id tickets: %w", err)
+	}
+
+	return tickets, nil
+}
+
+// RegTicket implements pastel.Client.RegTicket
+func (client *client) RegTicket(ctx context.Context, regTxid string) (*RegTicket, error) {
+	ticket := &RegTicket{}
```
alonlong, line 118: i think here should be: ticket := RegTicket{}

pastel/client.go
```diff
@@ -102,6 +102,28 @@ func (client *client) Verify(ctx context.Context, data []byte, signature, pastel
...
+
+	if err := client.callFor(ctx, &ticket, "tickets", "get", regTxid); err != nil {
+		return nil, errors.Errorf("failed to get reg ticket: %w", err)
+	}
+
+	return ticket, nil
+}
+
 func (client *client) callFor(ctx context.Context, object interface{}, method string, params ...interface{}) error {
```
alonlong, line 127: This function makes me a bit confused, in my opinion, i think when call the callFor function, the seconds parameter should be a pointer, then for CallForWithContext, just need to pass the object itself, not a &object, i know maybe u want to check it the remote call has the reply, but i think it’s not a good idea to do like this, take the grpc method for example, we won’t check if the reply is nil. And i can see the cases for call the callFor are not consistent

mateeullahmalik: @alonlong  I didn't write this function so probably its not right for me to change what's in master. But happy to make the changes if @levkoburburas agrees with you. 
Personally, I think we cannot set parameter as something like *interface{} as its not allowed. And someone calling this function would have to "guess" that they need to pass pointer to an object which is risky.

pastel/pastel.go
```diff
@@ -33,4 +33,10 @@ type Client interface {
 	// Verify verifies signed data by the given its signature and pastelID, if successful returns true.
 	// Command `pastelid verify "text" "signature" "PastelID"`.
 	Verify(ctx context.Context, data []byte, signature, pastelID string) (ok bool, err error)
+	// ActTickets returns activated art tickets.
+	// Command `tickets list act`.
+	ActTickets(ctx context.Context, actType ActTicketType, minHeight int) (ActTickets, error)
+	// ActTickets returns art registration tickets.
+	// Command `tickets get <txid>`.
+	RegTicket(ctx context.Context, regTxid string) (RegTicket, error)
```
pjenys0, line 41: nit: can add action prefix in these API? this can help prevent confusing when code volume get bigger.
like:
- QueryActTickets()
- GetRegTicket()

pastel/reg_ticket.go
```diff
@@ -0,0 +1,71 @@
...
+	ArtTicket     []byte    `json:"art_ticket"`
+	ArtTicketData ArtTicket `json:"-"`
+}
+
+// ArtTicket is Pastel Art Ticket
+type ArtTicket struct {
+	Version       int       `json:"version"`
+	Blocknum      int       `json:"blocknum"`
+	Author        []byte    `json:"author"`
+	DataHash      []byte    `json:"data_hash"`
```
pjenys0, line 33: check http://pastel.wiki/en/Architecture/Components/TicketStructures then this should be `block_hash`.
Could you confirm?

pastel/pastel.go
mateeullahmalik: 1- @levkoburburas  suggested on naming ActivationTickets as ActTicket and keeping it consistent throughout the project. 
2- Also suggested to avoid words like `GET`  as prefixes to function names

we only have one Activation ticket & Registration ticket... AFAIK

walletnode/cmd/app.go
```diff
@@ -115,17 +119,22 @@ func runApp(ctx context.Context, config *configs.Config) error {
 	// entities
 	pastelClient := pastel.NewClient(config.Pastel)
 	nodeClient := grpc.NewClient()
+	// p2p service (currently using kademlia)
```
pjenys0, line 122: nit: duplicate?

walletnode/cmd/app.go
```diff
@@ -26,6 +29,7 @@ import (
 const (
 	appName  = "walletnode"
 	appUsage = "WalletNode" // TODO: Write a clear description.
+	p2pDir   = "./p2p"      // relatively from work-dir
```
pjenys0, line 32: not sure where to use this?

walletnode/services/artworksearch/task.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Error returns task err
+func (task *Task) Error() error {
+	return task.err
+}
+
+// SubscribeSearchResult returns a new search resultof the state.
+func (task *Task) SubscribeSearchResult() <-chan *RegTicketSearch {
+
```
pjenys0, line 200: a little bit confusing here.
The Subscribe name raise a thought that allow multiple subscribers can listen the change.
But in this case we allow only 1 subscriber.
Anw. no big problem now.

pjenys0 (dismissed):
LGTM.

walletnode/cmd/app.go
mateeullahmalik: Thanks,
 removed this

walletnode/cmd/app.go
mateeullahmalik: good catch, 
removed.

pjenys0 (dismissed):
LGTM

pastel/pastel.go
```diff
@@ -33,4 +33,10 @@ type Client interface {
 	// Verify verifies signed data by the given its signature and pastelID, if successful returns true.
 	// Command `pastelid verify "text" "signature" "PastelID"`.
 	Verify(ctx context.Context, data []byte, signature, pastelID string) (ok bool, err error)
+	// ActTickets returns activated art tickets.
+	// Command `tickets list act`.
+	ActTickets(ctx context.Context, actType ActTicketType, minHeight int) (ActTickets, error)
+	// ActTickets returns art registration tickets.
```
alonlong, line 39: And the comment here is not ActTickets

walletnode/api/services/artwork.go
```diff
@@ -149,11 +151,61 @@ func (service *Artwork) Mount(ctx context.Context, mux goahttp.Muxer) goahttp.Se
...
+	searchReq := fromArtSearchRequest(req)
+	taskID := service.search.AddTask(searchReq)
+	task := service.search.Task(taskID)
+
+	resultChan := task.SubscribeSearchResult()
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case srch, ok := <-resultChan:
```
alonlong, line 166: srch is search? if it's ok to use full name, please don't use the abbreviation, thank u

walletnode/services/artworksearch/helper.go
```diff
@@ -0,0 +1,31 @@
+package artworksearch
+
+import (
+	b64 "encoding/base64"
+	"encoding/json"
+	"fmt"
+)
+
+func inIntRange(val int, min *int, max *int) bool {
```
alonlong, line 9: please don't use pointer if no change for it in function

walletnode/services/artworksearch/helper_test.go
```diff
@@ -0,0 +1,79 @@
+package artworksearch
+
+import (
+	b64 "encoding/base64"
```
alonlong, line 4: base64 is conflicted with other packages of same name, if not, i think it's better not  rename the package name

walletnode/services/artworksearch/request.go
```diff
@@ -0,0 +1,63 @@
...
+	ArtSearchKeyword.String(),
+}
+
+// String returns the string val of ArtSearchQueryField
+func (f ArtSearchQueryField) String() string {
+	return string(f)
+}
+
+// ArtSearchRequest represents artwork search payload
+type ArtSearchRequest struct {
```
alonlong, line 30: i am just curious why this structure uses many pointer fields? it's necessary?

Dicklesworthstone (approved):

walletnode/services/artworksearch/helper.go
mateeullahmalik: these vals can be nil.. you can see nil checks inside func bodies. 
this is intended.

walletnode/services/artworksearch/helper_test.go
mateeullahmalik: @alonlong agree. but  this is how its done throughout in existing code I'm just keeping things consistent.

walletnode/services/artworksearch/request.go
mateeullahmalik: Yes. This is incoming request payload. the fields that correspond to parameters that are optional are defined as pointers. Because if a `param` is provided & we don't use pointers then that `param` will come in as `false` which will make it impossible to identify the true intention of user.

pastel/pastel.go
mateeullahmalik: Thanks for pointing out. Comment is ActTickets. Pls see line # 36
As you'd know, this is normal practice in Go when comments are muti-line

pjenys0 (approved):
LGTM!

Dicklesworthstone (approved):

PR #83, opened 2021-06-09
Move pastel utility repo to gonode

s4kibs4mi:
Moved `Pastel-Utility` from individual repo to this repo. Also refactored the code to make the standard same as supernode and walletnode.


tools/pastel-utility/cmd/init.go
```diff
@@ -0,0 +1,307 @@
...
+// Print success info log on successfully ran command, return error if fail
+func downloadZksnarkParams(ctx context.Context, path string, force bool) error {
+	log.WithContext(ctx).Infoln("Downloading zksnark files")
+
+	for _, zksnarkParamsName := range configs.ZksnarkParamsNames {
+		zksnarkParamsPath := filepath.Join(path, zksnarkParamsName)
+
+		if _, err := os.Stat(zksnarkParamsPath); os.IsNotExist(err) || force {
+			out, err := os.Create(zksnarkParamsPath)
+			if err != nil {
```
levkoburburas, line 130:
```suggestion
		if _, err := os.Stat(zksnarkParamsPath); err == nil && !force {
			log.WithContext(ctx).Infof("zksnark file %s already exists and force set to %v", zksnarkParamsPath, force)
			continue
		}
		
		out, err := os.Create(zksnarkParamsPath)
		if err != nil {
		....
```

Write simpler, no need to nest large chunks of code deeper.

tools/pastel-utility/cmd/init.go
```diff
@@ -0,0 +1,307 @@
...
+// checkLocalAndRouterFirewalls checks local and router firewalls and suggest what to open
+func checkLocalAndRouterFirewalls(ctx context.Context, requiredPorts []string) error {
+	baseURL := "https://portchecker.com?q=" + strings.Join(requiredPorts[:], ",")
+	// resp, err := http.Get(baseURL)
+	// if err != nil {
+	// 	log.WithContext(ctx).WithError(err).Errorf("Error requesting url\n")
+	// 	return errors.Errorf("Failed to request port url %v \n", err)
+	// }
+	// defer resp.Body.Close()
+	// ok := resp.StatusCode == http.StatusOK
```
levkoburburas, line 169: What to do with it? Delete or add comment  
```go
// TODO: ....
or 
// NOTE: ....
```

tools/pastel-utility/cmd/init.go
```diff
@@ -0,0 +1,307 @@
...
+
+	log.WithContext(ctx).Infof("Config: ")
+	log.WithContext(ctx).Infof(config.String())
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	sys.RegisterInterruptHandler(cancel, func() {
+		log.WithContext(ctx).Info("Interrupt signal received. Gracefully shutting down...")
+		os.Exit(0)
```
levkoburburas, line 237: Again, It looks like this was done on purpose.
Have you ever heard about `<-ctx.Done()` ?

tools/pastel-utility/cmd/init.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+	}
+
+	v := viper.New()
+	v.SetConfigType("json")
+	if err := v.ReadConfig(bytes.NewBufferString(config.Init.String())); err != nil {
+		return err
+	}
+
+	walletConfigPath := fmt.Sprintf("%s/wallet.env", walletNodePath)
```
levkoburburas, line 209:
```suggestion
	walletConfigPath := filepath.Join(walletNodePath, "walletnode.yml")
```
1. Do you realize that windows OS has `\` back slash? Use `fmt.Sprintf` to concat paths is silly
2. `wallet.env` this is wrong filename! Who did tell you that config name of `walletnode` is `wallet.env` ? Moreover you are use `v.SetConfigType("json")`. It's weird 
3. Config is not needed to be placed in `work-dir`

tools/pastel-utility/cmd/init.go
```diff
@@ -0,0 +1,307 @@
...
+	log.WithContext(ctx).Infof("Config: ")
+	log.WithContext(ctx).Infof(config.String())
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	sys.RegisterInterruptHandler(cancel, func() {
+		log.WithContext(ctx).Info("Interrupt signal received. Gracefully shutting down...")
+		os.Exit(0)
+	})
```
levkoburburas, line 193: Why do you repeat it in each function? Is there no other way to write once?

levkoburburas (changes requested):
I only partly reviewed `cmd/init.go`. 
The code is poorly thought out and contains a lot of silly things.

a-ok123:
outdated

PR #91, opened 2021-06-15
Kademlia refactor

alonlong:


Dicklesworthstone (dismissed):

p2p/kademlia/node.go
```diff
@@ -2,132 +2,99 @@ package kademlia
...
 
-// NewNetworkNode creates a new NetworkNode for bootstrapping
-func NewNetworkNode(ip string, port string) *NetworkNode {
-	p, _ := strconv.Atoi(port)
-	return &NetworkNode{
-		IP:   net.ParseIP(ip),
-		Port: p,
+// NewNode returns a new node for bootstrapping
+func NewNode(ip string, port int) *Node {
+	return &Node{
```
pjenys0, line 30: where to set ID of node?

p2p/kademlia/network.go
```diff
@@ -0,0 +1,286 @@
...
+			log.WithContext(ctx).Errorf("conn write: %v", err)
+		}
+	}
+}
+
+// serve the incomming connection
+func (s *Network) serve(ctx context.Context) {
+	for {
+		// accept the incomming connections
+		conn, err := s.socket.Accept()
```
pjenys0, line 240: should also monitor ctx.Done() here to stop serving?

p2p/kademlia/network.go
```diff
@@ -0,0 +1,286 @@
...
+		case StoreData:
+			// handle the request for storing data
+			encoded, err := s.handleStoreData(ctx, request)
+			if err != nil {
+				log.WithContext(ctx).Errorf("handle store data request: %v", err)
+				continue
+			}
+			response = encoded
+		default:
+			log.WithContext(ctx).Errorf("impossible: invalid message type: %v", request.MessageType)
```
pjenys0, line 226: should return here to exist

p2p/kademlia/network.go
```diff
@@ -0,0 +1,286 @@
...
+func (s *Network) serve(ctx context.Context) {
+	for {
+		// accept the incomming connections
+		conn, err := s.socket.Accept()
+		if err != nil {
+			if strings.Contains(err.Error(), "closed") {
+				return
+			}
+
+			log.WithContext(ctx).Errorf("socket accept: %v", err)
```
pjenys0, line 246: in case of socket error (maybe due to network error), if we stop here then we will lost connection to other nodes.
Do you think should try to restart network again?

p2p/kademlia/network.go
```diff
@@ -0,0 +1,286 @@
...
+			return
+		}
+		log.WithContext(ctx).Debugf("%v: incomming connection: %v", s.self.String(), conn.RemoteAddr())
+
+		// handle the connection requests
+		go s.handleConn(ctx, conn)
```
pjenys0, line 252: I think should have limitation of number of concurrent connections

p2p/kademlia/hashtable.go
```diff
@@ -2,155 +2,205 @@ package kademlia
...
 	// Nodes within buckets are sorted by least recently seen e.g.
 	// [ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ][ ]
 	//  ^                                                           ^
 	//  └ Least recently seen                    Most recently seen ┘
-	RoutingTable [][]*node // 160x20
+	routeTable [][]*Node // 160x20
 
-	mutex *sync.Mutex
+	// mutex for route table
+	mutex sync.Mutex
```
pjenys0, line 46: can use RWMutex?

p2p/kademlia/network.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+}
+
+// Call sends the request to target and receive the response
+func (s *Network) Call(ctx context.Context, request *Message) (*Message, error) {
+	remoteAddr := fmt.Sprintf("%s:%d", request.Sender.IP, request.Receiver.Port)
+
+	// dial the remote address with udp network
+	conn, err := utp.DialContext(ctx, remoteAddr)
```
pjenys0, line 261: I think we should have a limited number of peers.
If we usually have to sync data among peers, then should keep connections alive in a connection table then create new connection for each requests.
How do you think?

p2p/kademlia/store/mem/mem.go
```diff
@@ -0,0 +1,92 @@
...
+	"sync"
+	"time"
+
+	"github.com/jbenet/go-base58"
+	"github.com/pastelnetwork/gonode/common/log"
+)
+
+// Store is a simple in-memory key/value store used for unit testing
+type Store struct {
+	mutex        sync.Mutex
```
pjenys0, line 14: can use RWMutex even this for testing :)

p2p/kademlia/dht.go
```diff
@@ -35,661 +39,436 @@ type Options struct {
...
-		return dht.timers(ctx)
-	})
-	group.Go(func() error {
-		return dht.networking.listen(ctx)
-	})
-	return group.Wait()
+	// retrieve the key/value from local storage
+	value, err := s.store.Retrieve(ctx, decoded)
+	if err != nil {
+		return nil, fmt.Errorf("store retrieve: %v", err)
```
pjenys0, line 171: if error happen get from local storage(like data corruption), do you think should try get from another peer, to prevent user get stuck here?
certainly should have a way to monitor health of local storage.

p2p/kademlia/dht.go
```diff
@@ -35,661 +39,436 @@ type Options struct {
...
+		}
+		contacted[string(node.ID)] = true
+
+		// update the running goroutines
+		number++
+
+		log.WithContext(ctx).Debugf("start work %v for node(%d): %s", iterativeType, len(nl.Nodes), node.String())
+
+		wg.Add(1)
+		// send and recive message concurrently
```
pjenys0, line 273: seem that we have to wait all responses from all peers before do processing responses.
So could we can change it to like:
```
func (s *DHT) doMultiWorkers(ctx,...) chan *Message {
	responseChan := make(chan * Message)
	go {
		var wg sync.WaitGroup
		// send the message to the first (closest) alpha nodes
		for _, node := range nl.Nodes {
			....
			wg.Add(1)
			// send and recive message concurrently
			go func(receiver *Node) {
				defer wg.Done()
				...
				...
				responseChan <- newMessage 
			}
		}

		wg.Wait()
		close(responseChan)
	}
}


for message := doMultiWorkers(ctx,...) {
	
}
```

p2p/kademlia/node.go
alonlong: this function is only used for bootstrap, i think it's better to remove it

p2p/kademlia/network.go
alonlong: i think it's better to monitor the done in err process

p2p/kademlia/network.go
alonlong: done

p2p/kademlia/network.go
alonlong: Agree, i add the logic according to the http server's accept

p2p/kademlia/network.go
alonlong: no, i don't think so

p2p/kademlia/hashtable.go
alonlong: of course, we can use RLock for the read operation

p2p/kademlia/network.go
alonlong: Yeah, it's an optimization for this part, but i think it's better to do it in the future.

p2p/kademlia/dht.go
alonlong: Good idea

p2p/kademlia/dht.go
alonlong: according to the whole logic, it's few different between the two ways, but i like the current way, i think it's more clean

a-ok123 (approved):

p2p/kademlia/network.go
pjenys0: I dont' think so, but anw, we can support later.

p2p/kademlia/network.go
pjenys0: sure. Thanks

p2p/kademlia/dht.go
pjenys0: That is different : for your one : you wait for all responses available before processing.
I suggest handle any response when it's available. That can have better performance, like in case poor connectivity

pjenys0 (approved):
It's fine to me now.
but still have some concern of improvement. put it in slack for further discussion.

pjenys0 (approved):
LGTM. Thanks @alonlong.

p2p/kademlia/dht.go
alonlong: done

Dicklesworthstone (approved):

PR #100, opened 2021-06-26
add two thumbnail apis for supernode

alonlong:


supernode/cmd/app.go
```diff
@@ -79,6 +81,37 @@ func NewApp() *cli.App {
 			}
 		}
 
+		// fetch the master node config by pastel client
+		pastelClient := pastel.NewClient(config.Pastel)
+		masterConfig, err := pastelClient.MasterNodeConfig(ctx)
+		if err == nil {
```
pjenys0, line 87: could you log error in case of error?
also,

proto/walletnode/register_artwork.proto
```diff
@@ -65,3 +71,23 @@ message SendTicketReply {
     string ticket_txid = 1;
 }
 
+message Thumbnail {
+    string key = 1;
+    bytes small = 2;
+    bytes medium = 3;
+    bytes large = 4;
+}
```
pjenys0, line 79: can include metadata of each thumbnail (like size?)

supernode/node/grpc/server/services/walletnode/register_artwork.go
```diff
@@ -178,6 +178,51 @@ func (service *RegisterArtwork) ProbeImage(stream pb.RegisterArtwork_ProbeImageS
...
+// AddThumbnail implements walletnode.RegisterArtworkServer.AddThumbnail()
+func (service *RegisterArtwork) AddThumbnail(ctx context.Context, req *pb.AddThumbnailRequest) (*pb.AddThumbnailReply, error) {
+	if req.Thumbnail == nil {
+		return nil, errors.New("thumbnail is empty")
+	}
+
+	log.WithContext(ctx).WithField("req", req).Debugf("AddThumbnail request")
+	task, err := service.TaskFromMD(ctx)
+	if err != nil {
+		return nil, err
```
pjenys0, line 190: could you wrap error before return?
like  errors.Errorf("get task: %w", error)

supernode/node/grpc/server/services/walletnode/register_artwork.go
```diff
@@ -178,6 +178,51 @@ func (service *RegisterArtwork) ProbeImage(stream pb.RegisterArtwork_ProbeImageS
...
+// GetThumbnail implements walletnode.RegisterArtworkServer.GetThumbnail()
+func (service *RegisterArtwork) GetThumbnail(ctx context.Context, req *pb.GetThumbnailRequest) (*pb.GetThumbnailReply, error) {
+	if req.Key == "" {
+		return nil, errors.New("key is empty")
+	}
+
+	log.WithContext(ctx).WithField("req", req).Debugf("GetThumbnail request")
+	task, err := service.TaskFromMD(ctx)
+	if err != nil {
+		return nil, err
```
pjenys0, line 211: same above, should wrap error before return.

metadb/client.go
```diff
@@ -0,0 +1,11 @@
+package metadb
+
+import "context"
+
+// Client exposes the interfaces for metadb service
+type Client interface {
+	// Query execute a query, not support multple statements
```
pjenys0, line 7: I think this metaDb is quite generic, not specified for our case.
For our case, I don't think we need a complex capacity of full SQL.
Simply Write(key, value)/Read(key, value) is fine for us.
and value will contains every metadata associate with key.

And for more generic, we can have like Write(ctx, namespace, key, value)/ Read(ctx, namespace, key, value).
Namespace can be like type, table,..
How do you think?

pjenys0: or more specific, metadb can provide service specific to our system like:
AddThumbnail(id, &Thumbnail{small, medium,large})
GetThumbnail(id) -> *Thumbnail
AddMetadata(id)..
GetMetadata(id)...

And details of implementation can be easy to map to any kind of database(rqlite in our case), like create Thumbnail table should be inside code of metadb than afterFun is set up in AfterFunc()

pjenys0:
Because metadb layer is in development, the way to store thumbnail iff changed. close it.

PR #103, opened 2021-06-30
Define some interfaces to pastel cNode, raptorq service, wallnode-supernode

pjenys0:
- Define some interfaces between wallnode-supernode, pastel interfaces
- Add raptorq service interface + test


raptorq/raptorq.proto
```diff
@@ -0,0 +1,73 @@
...
+  bytes data = 1;
+}
+
+message EncoderInfoReply {
+  repeated string name = 1;
+  EncoderParameters encoder_params = 2;
+}
+
+message EncoderParameters {
+  uint64 transfer_length = 1;   //u64
```
leonidcode20, line 53: why uint32 but the comment is //u16 ?

leonidcode20 (approved):
LGTM but please correct the comment in proto file which may confuse reader.

PR #105, opened 2021-07-01
refactor thumbnail fetch to use supernode & rqlite

mateeullahmalik:
- add code infra to spawn 10 connections with supernode & fetch thumbnails in parallel and return result back to the caller as soon as a thumbnail is received.
- added unit test cases of this new code infra & refactored previously existing tests to cope up with new changes - coverage of artsearch package now 80.5 % 
- This requires THumbnail API which is under progress under a different [PR](https://github.com/pastelnetwork/gonode/pull/105)
 So right now a place holder is added until that API is available.

### DIFF
--- a/walletnode/services/artworkregister/config.go
+++ b/walletnode/services/artworkregister/config.go
@@ -25,7 +25,6 @@ type Config struct {
 	// internal settings
 	connectToNextNodeDelay time.Duration
 	acceptNodesTimeout     time.Duration
-	connectTimeout         time.Duration
 
 	thumbnailSize int
 }
@@ -37,7 +36,6 @@ func NewConfig() *Config {
 
 		connectToNextNodeDelay: connectToNextNodeDelay,
 		acceptNodesTimeout:     acceptNodesTimeout,
-		connectTimeout:         connectTimeout,
 
 		thumbnailSize: thumbnailSize,
 	}
--- a/walletnode/services/artworksearch/thumbnail/thumbnail.go
+++ b/walletnode/services/artworksearch/thumbnail/thumbnail.go
@@ -1,0 +1,163 @@
+package thumbnail
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/pastelnetwork/gonode/common/log"
+	"github.com/pastelnetwork/gonode/pastel"
+	nodeClient "github.com/pastelnetwork/gonode/walletnode/node"
+	"github.com/pastelnetwork/gonode/walletnode/services/artworkregister/node"
+)
+
+const (
+	maxConnections    = 10
+	maxConnectionsErr = "cannot connect to more than 10 nodes"
+)
+
+// Helper is interface contract for Thumbnail Getter
+type Helper interface {
+	Connect(ctx context.Context, connections uint) error
+	Fetch(ctx context.Context, key string) ([]byte, error)
+	Close()
+}
+
+type thumbnailHelper struct {
+	pastelClient pastel.Client
+	nodeClient   nodeClient.Client
+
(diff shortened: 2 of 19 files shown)


walletnode/services/artworksearch/thumbnail/thumbnail.go
```diff
@@ -0,0 +1,163 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/pastelnetwork/gonode/common/log"
+	"github.com/pastelnetwork/gonode/pastel"
+	nodeClient "github.com/pastelnetwork/gonode/walletnode/node"
+	"github.com/pastelnetwork/gonode/walletnode/services/artworkregister/node"
```
pjenys0, line 13: you refer to same client interface to artwork_register to connect to supernode. Make me confusing.
Suggest define own gprc proto for artwork_search service in ./proto/walletnode. Even it's empty now.

walletnode/services/artworksearch/thumbnail/thumbnail.go
```diff
@@ -0,0 +1,163 @@
...
+
+	nodes, err := t.pastelTopNodes(ctx, connections)
+	if err != nil {
+		return fmt.Errorf("pastelTopNodes: %v", err)
+	}
+
+	for _, node := range nodes {
+		node := node
+		if err := node.Connect(ctx, t.timeOut); err != nil {
+			return err
```
pjenys0, line 72: you return if any connect error - but leave some previous connect success without closing it.
I think we can consider two options :
1. if any error - stop all listening goroutine
2. return error if all connect() return error, and return ok if any connect got success.

walletnode/services/artworksearch/thumbnail/thumbnail.go
```diff
@@ -0,0 +1,163 @@
...
+
+		nodes = append(nodes, node.NewNode(t.nodeClient, mn.ExtAddress, mn.ExtKey))
+	}
+
+	return nodes, nil
+}
+
+// Close closes request channel. Once Close is called, the helper can no longer fetch thumbnails
+func (t *thumbnailHelper) Close() {
+	if t.isClosed {
```
pjenys0, line 146: why this check is outside of mutex locking section?

mateeullahmalik:
Update: resolved @pjenys0 comments & fixed breaking test cases due to new code addition

walletnode/services/artworksearch/thumbnail/thumbnail.go
mateeullahmalik: fixed, good catch!

walletnode/services/artworksearch/thumbnail/thumbnail.go
mateeullahmalik: applied (1) I think that makes more sense

walletnode/services/artworksearch/thumbnail/thumbnail.go
mateeullahmalik: done

walletnode/services/artworkregister/config.go
```diff
@@ -25,7 +25,6 @@ type Config struct {
 	// internal settings
 	connectToNextNodeDelay time.Duration
 	acceptNodesTimeout     time.Duration
-	connectTimeout         time.Duration
```
leonidcode20, line 28: Why do we remove `connectTimeout` here ?

mateeullahmalik: this was added at a different place

leonidcode20 (approved):

pjenys0 (approved):

PR #107, opened 2021-07-03
Kademlia - Replace sha1 with sha3-256.

danileng20:


alonlong (approved):

p2p/kademlia/hashtable.go
```diff
@@ -23,8 +23,8 @@ const (
 	Alpha = 3
 
 	// B - the size in bits of the keys used to identify nodes and store and
-	// retrieve data; in basic Kademlia this is 160, the length of a SHA1
-	B = 160
+	// retrieve data; in basic Kademlia this is 256, the length of a SHA3-256
+	B = 256
```
pjenys0, line 27: could you improve name of these constants? simply B has no meaning

danileng20: I think it's better if we let it be.
These constants are defined it the Kademlia's specs. http://xlattice.sourceforge.net/components/protocol/kademlia/specs.html

pjenys0 (approved):

leonidcode20 (approved):
LGTM!

PR #111, opened 2021-07-06
[Draft] Art reg -  add more pastel interfaces

pjenys0:
TBD

### DIFF
--- a/pastel/client.go
+++ b/pastel/client.go
@@ -136,7 +136,7 @@ func (client *client) RegTicket(ctx context.Context, regTxid string) (RegTicket,
 	return ticket, nil
 }
 
-func (client *client) GetBlockCount(ctx context.Context) (int64, error) {
+func (client *client) GetBlockCount(ctx context.Context) (int32, error) {
 	res, err := client.CallWithContext(ctx, "getblockcount", "")
 	if err != nil {
 		return 0, errors.Errorf("failed to call getblockcount: %w", err)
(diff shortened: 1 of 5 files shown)


pastel/client.go
```diff
@@ -146,7 +156,88 @@ func (client *client) GetBlockCount(ctx context.Context) (int64, error) {
...
+	return artticketFee.ArtticketFee, nil
+}
+
+func (client *client) GetRegisterArtFee(ctx context.Context, request GetRegisterArtFeeRequest) (int64, error) {
+	var totalStorageFee struct {
+		TotalStorageFee int64 `json:"totalstoragefee"`
+	}
+
+	// FIXME: fix here
+	if err := client.callFor(ctx, &totalStorageFee, "gettotalstoragefee", "TBD"); err != nil {
```
pjenys0, line 225: I'm filling GetRegisterArtFeeRequest into list of parameters, will push later.

pastel/client.go
```diff
@@ -136,7 +136,17 @@ func (client *client) RegTicket(ctx context.Context, regTxid string) (RegTicket,
...
+	result := &GetBlockVerbose1Result{}
+
+	if err := client.callFor(ctx, result, "getblock", fmt.Sprint(blkHeight), "1"); err != nil {
+		return result, errors.Errorf("failed to get block: %w", err)
+	}
+
+	return result, nil
+}
+
+func (client *client) GetBlockCount(ctx context.Context) (int32, error) {
```
leonidcode20, line 149: Return type should be int64 because `res.GetInt()` return `int64`. We may risk truncate value here.

PR #116, opened 2021-07-12
add config for rqfiles dir in wn & sn

mateeullahmalik:


mykytaccode0 (changes requested):
Missing the code to create the directory on the machine, need to do similar to:
		if err := os.MkdirAll(config.TempDir, os.ModePerm); err != nil {
			return errors.Errorf("could not create temp-dir %q, %w", config.TempDir, err)
		}

mykytaccode0 (approved):

walletnode/cmd/app.go
```diff
@@ -95,6 +98,10 @@ func NewApp() *cli.App {
 			return errors.Errorf("could not create temp-dir %q, %w", config.TempDir, err)
 		}
 
+		if err := os.MkdirAll(config.RqFilesDir, os.ModePerm); err != nil {
```
pjenys0, line 101: could you put directly `0777` than os.ModePerm? Because os.ModePerm is usually used as bit mask than directly.

pjenys0 (approved):
LGTM!

PR #117, opened 2021-07-14
add test cases for artwork register svc

mateeullahmalik:


walletnode/services/artworkregister/rqids.go
```diff
@@ -16,11 +16,11 @@ type RQIDS struct {
 	Content []byte
 }
 
-func (rqidsList RQIDSList) Identifiers() []string {
-	var identifier []string
+func (rqidsList RQIDSList) Identifiers() (identifier []string) {
```
leonidcode20, line 19: nit: name returns is not necessary here and if you use name return then `identifier` at line `24` is redundant

mateeullahmalik: @leonidcode20 fixed

leonidcode20 (approved):

pjenys0 (approved):
Thanks for support!

PR #121, opened 2021-07-18
add artreg flow test cases

mateeullahmalik:
- came across a bug & fixed it: The setPrimary function of nodes List was causing panic because
   (a) it wasn't checking whether that index actually exists in the list.
   (b) on the calling function side, the caller has no straight forward way of knowing it as well. It was actually breaking the code                            inside meshNodes function Line 453: `meshNodes.SetPrimary(primaryIndex)` here, meshNodes is a list of node & this `primary index` int was non-existent in the list which was causing panic.
   (c) It makes more sense to tie the function to the node object as the node object is supposed to be primary.. not the nodes list.
   
- fixed test cases that were breaking due to new code addition in walletnode  & supernode 
- added new test cases in walletnode artwork register service


common/service/artwork/file.go
```diff
@@ -24,6 +24,7 @@ type File struct {
 	fmt.Stringer
 	sync.Mutex
 
+	storage.File
```
pjenys0, line 27: while need it? not safe to do it now. easy to miss-use it.

walletnode/services/artworkregister/helper.go
```diff
@@ -4,9 +4,22 @@ import (
 	"bytes"
 	"io"
 
+	"github.com/pastelnetwork/gonode/common/errors"
 	"golang.org/x/crypto/sha3"
 )
 
+var (
```
leonidcode20, line 11: Really like this. Thank you

PR #128, opened 2021-07-29
fix waitTxnValid bugs

mateeullahmalik:
Switch case will never run timeout branch. I identified & double-checked it while writing test cases. 
Please see this  [stack overflow post](https://stackoverflow.com/questions/23582143/golang-using-timeouts-with-channels)
Fixed it & also made timeout an argument as test cases can't afford 15 seconds of wait where not needed.

### DIFF
--- a/walletnode/services/artworkregister/task.go
+++ b/walletnode/services/artworkregister/task.go
@@ -158,13 +158,13 @@ func (task *Task) run(ctx context.Context) error {
 	close(nodesDone)
 	for i := range task.nodes {
 		if err := task.nodes[i].Connection.Close(); err != nil {
-			log.WithContext(ctx).Debugf("failed to close connection to node %s %s", task.nodes[i].PastelID(), err)
+			log.WithContext(ctx).WithError(err).Debugf("failed to close connection to node %s", task.nodes[i].PastelID())
 		}
 	}
 
 	// new context because the old context already cancelled
 	newCtx := context.Background()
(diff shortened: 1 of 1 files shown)


walletnode/services/artworkregister/task.go
```diff
@@ -213,7 +221,7 @@ func (task *Task) waitTxidValid(ctx context.Context, txID string, expectedConfir
 			} else {
 				return nil
 			}
-		case <-time.After(timeout):
+		case <-timeoutCh:
```
leonidcode20, line 224: Thank you. Good catch here because `time.After` is rearmed every time the select run again.

walletnode/services/artworkregister/task.go
```diff
@@ -185,19 +185,27 @@ func (task *Task) run(ctx context.Context) error {
...
 }
 
-func (task *Task) waitTxidValid(ctx context.Context, txID string, expectedConfirms int64, timeout time.Duration) error {
+func (task *Task) waitTxidValid(ctx context.Context, txID string, expectedConfirms int64, timeout time.Duration, interval time.Duration) error {
+	checkTimeout := func(checked chan<- bool) {
+		time.Sleep(timeout)
+		checked <- true
+	}
+
+	timeoutCh := make(chan bool)
```
leonidcode20, line 194: Just two things:
1. `chan struct{}` is better.
2.  `close(checked)` is better in my opinion.

mateeullahmalik: Thank you! Changes made

PR #131, opened 2021-07-31
add artreg flow test cases

mateeullahmalik:
added new test cases
please load diff of file `walletnode/services/artworkregister/task_test.go`
its not loaded by default due to large diff

### DIFF
--- a/walletnode/services/artworkregister/node/node.go
+++ b/walletnode/services/artworkregister/node/node.go
@@ -59,6 +59,16 @@ func (node *Node) Connect(ctx context.Context, timeout time.Duration) error {
 	return nil
 }
 
+// SetPrimary promotes a supernode to primary role which handle the write to Kamedila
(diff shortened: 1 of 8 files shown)


walletnode/services/artworkregister/node/node.go
```diff
@@ -49,7 +50,9 @@ func (node *Node) Connect(ctx context.Context, timeout time.Duration) error {
 
 	connCtx, connCancel := context.WithTimeout(ctx, timeout)
 	defer connCancel()
-
+	if node.Client == nil {
+		fmt.Println("im ni")
```
pjenys0, line 54: should return error.

mateeullahmalik: It was a typo. removed. Thanks for the catch

pjenys0 (approved):
LGTM. Thanks @matee81

leonidcode20 (approved):

PR #133, opened 2021-08-01
Dupedection interact

danileng20:
This PR is to add a dupe detection API that interacts with dupe detection service(https://github.com/pastelnetwork/dd-service) to get fingerprints, rareness scores and other info.
Task: https://freedcamp.com/view/2956207/tasks/41229484


dupedetection/client.go
```diff
@@ -0,0 +1,127 @@
...
+	return outputBase, outputPath, nil
+}
+
+func (client *client) collectOutput(ctx context.Context, baseName, path string) (*DupeDetection, error) {
+	// Waiting for JSON output from dupe detection service
+	ticker := time.NewTicker(1 * time.Second)
+	for {
+		select {
+		case <-ctx.Done():
+			ticker.Stop()
```
leonidcode20, line 70: should be `defer ticker.Stop()` because other branch doesn't call it.

dupedetection/client.go
```diff
@@ -0,0 +1,127 @@
...
+			return nil, errors.New("Timeout")
```
pjenys0, line 71: error string should not start with upper case

dupedetection/client.go
```diff
@@ -0,0 +1,127 @@
...
+	// Collect result from dupe detection output directory
+	result, err := client.collectOutput(ctx, outputBase, outputPath)
+	if err != nil {
+		err = errors.Errorf("failed to collect output from dupe detection service: %w", err)
+	}
+
+	return result, err
+}
+
+func (client *client) copyImageToInputDir(inputPath string) (string, string, error) {
```
leonidcode20, line 38: If we already agree on the rule how "output" file will be named. For example, output = "input.json" then return the expected output file name is more efficient than  walking the directory.

pjenys0 (approved):
Look good to me - excepts:
- small comment related to error
- also consider clear input/output file after finish generation to save memory

dupedetection/client.go
```diff
@@ -0,0 +1,127 @@
...
+					}
+
+					return result, nil
+				}
+			}
+		}
+	}
+}
+
+func parseOutput(path string) (*DupeDetection, error) {
```
leonidcode20, line 93: Should have a timeout here because if dd-service is crashed/restarted and doesn't pick up left-over files then this thread will leak/hang until caller abort.

dupedetection/client.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	outputBase := inputName + ".json"
+	outputPath := filepath.Join(client.config.OutputDir, outputBase)
+	return outputBase, outputPath, nil
+}
+
+func (client *client) collectOutput(ctx context.Context, baseName, path string) (*DupeDetection, error) {
+	// Waiting for JSON output from dupe detection service
+	ticker := time.NewTicker(1 * time.Second)
```
georgkcode0, line 66: Better if we define the duration time in config (it can get from config file or by default)

dupedetection/client.go
danileng20: I also considered this point. But not sure about how long it takes from dd-service side, so I decided to let the caller stop it via the context of the caller.

dupedetection/client.go
danileng20: fixed. Thanks.

dupedetection/client.go
danileng20: fixed. Thanks.

dupedetection/client.go
danileng20: This point is very small, and we will not touch it in the future, so I decided to define it here. Thanks.

leonidcode20 (approved):

dupedetection/client.go
danileng20: okay, I added a timeout.

pjenys0 (approved):

leonidcode20 (approved):

PR #144, opened 2021-08-09
Fix wrong decoding issue of art-search

leonidcode20:
- Fix decode of reg-ticket in correctly.

### DIFF
--- a/walletnode/services/artworkregister/task.go
+++ b/walletnode/services/artworkregister/task.go
@@ -530,9 +530,9 @@ func (task *Task) createArtTicket(_ context.Context) error {
 	// TODO: fill all 0 and "TBD" value with real values when other API ready
 	ticket := &pastel.ArtTicket{
 		Version:   1,
-		Author:    pastelID,
+		Author:    string(pastelID),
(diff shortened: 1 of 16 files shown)


danileng20 (approved):

walletnode/services/artworkregister/task.go
```diff
@@ -530,9 +530,9 @@ func (task *Task) createArtTicket(_ context.Context) error {
 	// TODO: fill all 0 and "TBD" value with real values when other API ready
 	ticket := &pastel.ArtTicket{
 		Version:   1,
-		Author:    pastelID,
+		Author:    string(pastelID),
```
pjenys0, line 533: per https://pastel.wiki/en/Architecture/Components/TicketStructures then:
author and block_hash is bytes.
that is why above we converted ArtistPastelId(base58 format) to 
`pastelID := base58.Decode(task.Request.ArtistPastelID)`.

And you convert a byte to string is very dangerous here if bytes is not string.
You should using directly `task.Request.ArtistPastelID` if want this change. Same for `artistBlockHash`

PR #148, opened 2021-08-11
Fix decoding and search

mateeullahmalik:
fixes decoding issue of artTicket
fixes encoding issue
fixes breaking test cases, add new test case 
fix art-search bugs & issues revealed during testing
change rareness scores range to 0.0 - 1.0, updated API contratcs


pastel/reg_ticket.go
```diff
@@ -37,9 +37,9 @@ type RegTicketData struct {
 // ArtTicket is Pastel Art Ticket
 type ArtTicket struct {
 	Version       int       `json:"version"`
-	Author        []byte    `json:"author"`
+	Author        string    `json:"author"`
```
nvnguyen202, line 40: why this change is very important to you?

nvnguyen202: please check 
```
"version": integer          // 1
  "author": bytes,            // PastelID of the author (creator)
  "blocknum": integer,        // block number when the ticket was created - this is to map the ticket to the MNs that should process it
  "block_hash": bytes         // hash of the top block when the ticket was created - this is to map the ticket to the MNs that should process it
  "copies": integer,  
```

walletnode/services/artworkregister/task.go
```diff
@@ -530,9 +530,9 @@ func (task *Task) createArtTicket(_ context.Context) error {
 	// TODO: fill all 0 and "TBD" value with real values when other API ready
 	ticket := &pastel.ArtTicket{
 		Version:   1,
-		Author:    pastelID,
+		Author:    string(pastelID),
 		BlockNum:  task.artistBlockHeight,
-		BlockHash: task.artistBlockHash,
+		BlockHash: string(task.artistBlockHash),
```
nvnguyen202, line 535: this code is not safe.
`task.artistBlockHash, err = hex.DecodeString(blockInfo.Hash)` 
based on this then task.artistBlockHash is bytes, not sure is representation of a string. Casting it to string are not safe.

pastel/reg_ticket.go
mateeullahmalik: @nvnguyen202 This is necessary figured out after testing.

mateeullahmalik: I've tested the flow with cNode. changing it to []byte will break the unmarshal. Please see the screenshot below
image
image

walletnode/services/artworkregister/task.go
mateeullahmalik: Thanks for raising. As I mentioned above, I had to change these data types to string because unmarshal was failing on []byte, I don't think changing it to string is unsafe, we can go back and forth from string to []byte safely. please share if you have a reference. Also, If you think there is a better way, please let me know, I'll be happy to make changes. 
I know there is one way of declaring our new data type and writing a custom unmarshal function for it which does hex decoding but that will introduce ripple effects and I believe string would work fine. 
please share your  solution proposal.

pastel/reg_ticket.go
nvnguyen202: if so, let keep it. then try to test it.
but need to change:
```
// Decode hash string to byte
task.artistBlockHash, err = hex.DecodeString(blockInfo.Hash)
```
to :
```
1. change artistBlockHash -> to string type
2. task.artistBlockHash = blockInfo.Hash
```

mateeullahmalik: @nvnguyen202 I can happily change this one line but i spent half an hour on it and i think a different PR should do it after looking at things thoroughly because there is already a confusion around this... I came across this FIXME comment in the same file about same thing (block hash format)
probably someone who worked on artworkregister
or I could take a look as well but in the scope of a different PR. One reason is that the mock networks is sending hash which fails to hex decode. (will provide more info on this later). Updated mock networks will introduce other new updates which will break more things that we know about and plan to fix after merging current open PRs. 

image

walletnode/services/artworkregister/task.go
pjenys0: sorry, I wrong used my personal account.
You can convert a bytes to string - only if ensure bytes are encode of a correct string
because if not - you will meet some problem later. I saw it many times.
for example - you can try with follow program - and see how it returns:
```
package main

import (
	"fmt"
)

func main() {
	   a := []byte{0xff}
    s := string(a)
    fmt.Println(s)
    for _, r := range s {
        fmt.Println(r)
    }
    rs := []rune(s)
    fmt.Println(rs)
}

```

mateeullahmalik: @pjenys0  I'm not disagreeing with you. What' I'm saying is; we can unmarshal it into string only that's already established. 
after that, we have many options,..either we can hex decode it (use ecnodeArtticket & decodeArtticket functions) or some other way. Please read this:

https://stackoverflow.com/questions/65872616/go-illegal-base64-data-at-input-byte-0-when-unmarshalling-a-string-json

so, I do not disagree, I'm saying that the fix would be in the scope of a separate PR which only addresses this issue by following the entire flow.. because of the importance of this, it should be a separate PR. 
what do you think?

pjenys0: why need another PR?

pastel/reg_ticket.go
mateeullahmalik: @pjenys0 please read my comment above ^^ 
This is why we need a different PR to address this. 
changing this one line is not the solution here. we need to look into it thoroughly.

walletnode/services/artworkregister/task.go
pjenys0: if so, why you don't resolve only decode issue in this PR only?

mateeullahmalik: decode issue is resolved that's why APIs are working fine. I've attached screenshots of testing and working PRs above in the other thread, Please take a look.

mateeullahmalik: mock-networks in current state is returning wrong hashes which fails hex decode. 
If I update mock-networks now then other changes will come in as well (what ayk mentioned today morning) and that will break the entire thing again. 
we agreed on slack that we'll work on that once we merge these PRs

pastel/reg_ticket.go
```diff
@@ -106,7 +106,7 @@ type AppTicket struct {
 	AlternateNSFWScores     AlternateNSFWScores `json:"alternate_nsfw_scores"`
 	ImageHashes             ImageHashes         `json:"image_hashes"`
 
-	RQIDs []string `json:"rq_ids"`
+	RQIDs []string `json:"rqids"`
```
danileng20, line 109: Please refer to the document: https://pastel.wiki/en/Architecture/Components/TicketStructures
Why did you change here?

walletnode/services/artworkregister/task.go
pjenys0: anw. I don't want to take time on this now.
So please resolve below command of @danileng20 . Then I agree to your solution now. But - then please create PR to fix it later.
In mean while - please help to review https://github.com/pastelnetwork/gonode/pull/147

pastel/reg_ticket.go
mateeullahmalik: @danileng20 good catch. Thanks!

pjenys0 (dismissed):
LGTM, please create PR to resolve open items.

walletnode/services/artworkregister/task.go
mateeullahmalik: Sure, reviewing now

danileng20 (dismissed):

pjenys0 (dismissed):

danileng20 (dismissed):

danileng20 (approved):

pjenys0 (approved):

PR #150, opened 2021-08-12
Update logic Art Download with Trade ticket. And refactor + add debug log.

danileng20:
Update Art Download with Trade ticket.
Update file format for Art Reg at supernode.
Refactor + add debug log.


supernode/services/artworkregister/task.go
```diff
@@ -156,25 +162,30 @@ func (task *Task) ConnectTo(_ context.Context, nodeID, sessID string) error {
 		return err
 	}
 
+	var err error
+
 	task.NewAction(func(ctx context.Context) error {
 		node, err := task.pastelNodeByExtKey(ctx, nodeID)
```
pjenys0, line 168: could you declare node like:
```
var node *Node
node, err = ...
```
to prevent confusing with local scope

supernode/services/artworkregister/task.go
```diff
@@ -700,16 +740,22 @@ func (task *Task) UploadImageWithThumbnail(_ context.Context, file *artwork.File
 		// TODO: improve it by call stats on file
 		fileBytes, err := file.Bytes()
 		if err != nil {
-			return errors.Errorf("failed to read image file %w", err)
+			log.WithContext(ctx).WithError(err).Errorf("failed to read image file")
```
pjenys0, line 743: same here, can define:
```
var fileBytes []byte
fileBytes, err = file.Bytes()
```

supernode/services/artworkregister/task.go
danileng20: updated. Thanks.

supernode/services/artworkregister/task.go
danileng20: updated. Thanks.

mykytaccode0 (approved):

PR #157, opened 2021-08-16
Remove fingerprints comparison between original image and image with signature.

danileng20:
After discussing, we decided to remove fingerprints comparison between original image and image with signature.


leonidcode20 (approved):

pjenys0 (approved):
LGTM!

PR #159, opened 2021-08-18
fix art search match resp, thumbnails & testcases

mateeullahmalik:
- Fixes Thumbnail returns (return thumbnail 1 & thumbnail 2 in artsearch)
- Fixes testcases that were skipped earlier
- Fixes API response for match (so client can know the exact field that matches & the exact matching string in that field & the indexes of chars that match)


leonidcode20 (dismissed):
LGTM! But did you test wit the real cNode ?

walletnode/api/design/artworks.go
```diff
@@ -549,7 +549,11 @@ var ArtworkSummary = Type("ArtworkSummary", func() {
 	Description("Artwork response")
 
 	Attribute("thumbnail", Bytes, func() {
```
pjenys0, line 551: shoudl change `thumbnail` to `thumbnail_1` to unify with other parts.

pjenys0 (approved):
Thanks for update it. LGTM!

leonidcode20 (approved):

PR #162, opened 2021-08-20
Fix default pastel config path

mateeullahmalik:
- Fixes default pastel conf file path assigning
- SN & WN should not provide pastel configs in their own config files, rather, they should
    read pastel configs from pastel.conf
- Handle testnet param in pastel.conf
- Add unit test case


leonidcode20 (approved):
LGTM!

pjenys0 (approved):
LGTM

PR #173, opened 2021-08-28
Remove dependencies to external schema file

nguyen202:
Integrate schema content to go code than depend to external file.

### DIFF
--- a/metadb/database/service.go
+++ b/metadb/database/service.go
@@ -3,7 +3,6 @@ package database
 import (
 	"bytes"
 	"context"
-	"io/ioutil"
 	"strings"
 
 	"text/template"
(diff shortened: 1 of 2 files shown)


metadb/database/service.go
```diff
@@ -120,18 +119,16 @@ func (db *Ops) ReadUserData(ctx context.Context, artistPastelID string) (userdat
 
 // Run run the rqlite database service
 func (db *Ops) Run(ctx context.Context) error {
+	var err error
 
 	ctx = log.ContextWithPrefix(ctx, logPrefix)
 	log.WithContext(ctx).Info("start initialization")
 
-	content, err := ioutil.ReadFile(db.config.SchemaPath)
```
mateeullahmalik, line 127: Please remove this config var from config struct as well

mateeullahmalik (changes requested):
Please remove the unnecessary config var from  config struct

a-ok123 (approved):

metadb/database/service.go
nguyen202: @mateeullahmalik : let keep it. later maybe we will have bunch of template file.
we're thinking to support both of them - that should be easy for development later.

mateeullahmalik (approved):

PR #180, opened 2021-08-30
Update configs usage

mateeullahmalik:
- update sn configs
- update walletnode configs
- remove p2p from wn
- remove raptorQServiceAddr mapping
- remove bad defaults
- remove usage of metadb exposed addr 
- add input validations (empty checks) for p2p, metadb & raptorq configs

### DIFF
--- a/walletnode/examples/configs/localnet.yml
+++ b/walletnode/examples/configs/localnet.yml
@@ -5,6 +5,13 @@ temp-dir: "/tmp/walletnode"
 work-dir:
 rq-files-dir: 
 
+log-level:
+log-file:
+quiet: false
+temp-dir:
+work-dir:
+rq-files-dir: 
+
 node:
   api:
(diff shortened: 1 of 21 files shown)


walletnode/examples/configs/localnet.yml
```diff
@@ -5,29 +5,21 @@ temp-dir: "/tmp/walletnode"
 work-dir:
 rq-files-dir: 
 
+log-level:
+log-file:
+quiet: false
+temp-dir:
+work-dir:
+rq-files-dir:
```
a-ok123, line 13: these last two are in the lines 5 and 6 above too

mateeullahmalik: ah, good catch! Thank you

a-ok123 (approved):

nguyen202 (approved):

PR #182, opened 2021-08-31
Improve timeout config

nguyen202:
- Unify config name + default values
- Unify timeout values

### DIFF
--- a/raptorq/node/grpc/client.go
+++ b/raptorq/node/grpc/client.go
@@ -22,13 +22,13 @@ type client struct{}
 // Connect implements node.Client.Connect()
 func (client *client) Connect(ctx context.Context, address string) (node.Connection, error) {
 	// Limits the dial timeout, prevent got stuck too long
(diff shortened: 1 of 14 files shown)


raptorq/node/grpc/client.go
```diff
@@ -12,17 +13,22 @@ import (
...
+	defaultConnectTimeout = 5 * time.Second
 )
 
 type client struct{}
 
 // Connect implements node.Client.Connect()
 func (client *client) Connect(ctx context.Context, address string) (node.Connection, error) {
+	// Limits the dial timeout, prevent got stuck too long
+	dialCtx, cancel := context.WithTimeout(ctx, defaultConnectTimeout)
+	defer cancel()
```
nguyen202, line 26: should add a timeout context here to prevent got stuck here too long in case rqservice is not available.

mateeullahmalik (approved):

a-ok123 (approved):

PR #188, opened 2021-09-08
Active bootstap node

mateeullahmalik:
- update common/ keyvalue to use go-cache. It provides expiry support & fast in-memory cache
- add functionality to try re-seed boostrap node in case its unable to do so. It will check every 10 min till the time its successful to reach an active node. 
- in the meantime, it will keep bad addrs in a local cache and make sure not to use them for next 12 hrs 
- added test cases


p2p/kademlia/bootstrap.go
```diff
@@ -0,0 +1,154 @@
...
+
+	extP2p, err := get(ctx, s.pastelClient.MasterNodesTop)
+	if err != nil {
+		return fmt.Errorf("masternodesTop failed: %s", err)
+	} else if extP2p == "" {
+		extP2p, err = get(ctx, s.pastelClient.MasterNodesExtra)
+		if err != nil {
+			return fmt.Errorf("masternodesExtra failed: %s", err)
+		} else if extP2p == "" {
+			log.WithContext(ctx).Info("unable to fetch bootstrap ip. Missing extP2P")
```
nguyen202, line 51: should be Error() than Info()

p2p/kademlia/bootstrap.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil
+	}
+
+	var wg sync.WaitGroup
+	for _, node := range s.options.BootstrapNodes {
+		// sync the node id when it's empty
+		if len(node.ID) == 0 {
+			addr := fmt.Sprintf("%s:%v", node.IP, node.Port)
+			if _, err := s.cache.Get(addr); err == nil {
+				log.WithContext(ctx).WithField("addr", addr).Info("skip bad p2p boostrap addr")
```
nguyen202, line 97: if skip then should have `continue` here?

mateeullahmalik: great eyes! Thanks

nguyen202 (approved):
LGTM!

a-ok123 (approved):

PR #187, opened 2021-09-05
Add more information of system to ping service

nguyen202:
Add ping service to return healthcheck of system.

### New config params are introduced:
```
health-check:
  enable: true
  local-only: false
  port: 4445
  update-interval: "5s"
```

###  Sample for stats info:
```
{
    "metaDB": {
        "addr": "[::]:4042",
        "apply_timeout": "10s",
        "db_conf": {
            "DSN": "",
            "Memory": false
        },
        "dir": "/home/nvnguyen/.pastel/supernode/metadb-4444",
        "dir_size": 589824,
        "election_timeout": "1s",
        "heartbeat_timeout": "1s",
        "leader": {
            "addr": "[::]:4042",
            "node_id": "jXXsFY5CMSFy5T1cHFiqTsK2B5kMv93zP48G2FdfPWuPP6p2jH6c1JrNGL3VJHDxFvq6eZSsWppC3191sbG3oD"
        },
        "node_id": "jXXsFY5CMSFy5T1cHFiqTsK2B5kMv93zP48G2FdfPWuPP6p2jH6c1JrNGL3VJHDxFvq6eZSsWppC3191sbG3oD",
        "nodes": [
            {
                "id": "jXXsFY5CMSFy5T1cHFiqTsK2B5kMv93zP48G2FdfPWuPP6p2jH6c1JrNGL3VJHDxFvq6eZSsWppC3191sbG3oD",
                "addr": "[::]:4042",
                "suffrage": "Voter"
            }
        ],
        "raft": {
            "applied_index": 484,
            "commit_index": 484,
            "fsm_pending": 0,
            "last_contact": 0,
            "last_log_index": 484,
            "last_log_term": 36,
            "last_snapshot_index": 0,
            "last_snapshot_term": 0,
            "latest_configuration": "[{Suffrage:Voter ID:jXXsFY5CMSFy5T1cHFiqTsK2B5kMv93zP48G2FdfPWuPP6p2jH6c1JrNGL3VJHDxFvq6eZSsWppC3191sbG3oD Address:[::]:4042}]",
            "latest_configuration_index": 0,
            "log_size": 524288,
            "num_peers": 0,
            "protocol_version": 3,
            "protocol_version_max": 3,
            "protocol_version_min": 0,
            "snapshot_version_max": 1,
            "snapshot_version_min": 0,
            "state": "Leader",
            "term": 36
        },
        "request_marshaler": {
            "compression_batch": 5,
            "compression_size": 150,
            "force_compression": false
        },
[...]

PR #190, opened 2021-09-13
Add more p2p stats

nguyen202:
Add details of count of each kind of data in kademlia.
This depend on https://github.com/pastelnetwork/gonode/pull/187 - need merge it first


p2p/kademlia/dht.go
```diff
@@ -169,20 +192,43 @@ func (s *DHT) Store(ctx context.Context, data []byte) (string, error) {
...
 	if len(decoded) != B/8 {
 		return nil, fmt.Errorf("invalid key: %v", key)
 	}
 
+	actualKey := s.storedKey(keyType, decoded)
+
+	// retrieve the key/value from local storage
+	value, err := s.store.Retrieve(ctx, actualKey)
+	if err != nil {
+		log.WithContext(ctx).WithError(err).Error("store retrive failed")
```
mateeullahmalik, line 206: don't wanna return err?

mateeullahmalik (approved):

p2p/kademlia/dht.go
nguyen202: @matee81 : in case of error then value is nil, then we will try to request it from peers.

hoanlinh92 (approved):

PR #191, opened 2021-09-15
Fix zero config issue

nguyen202:
Some parameters are zero if not configured in config file


mateeullahmalik (approved):

mateeullahmalik:
@nguyen202 please double check that if config.var == 0 then what? I think zero values will still be assigned to let's say artworkregister.config.var? 
pls double check this.. I've approved the PR if you think these is fine.

nguyen202:
@matee81 : if config.var == 0, then preset default values used. I did test it.

hoanlinh92 (approved):

PR #196, opened 2021-09-25
Rqlite leader

mateeullahmalik:
Related ticket: https://freedcamp.com/view/2956207/tasks/42099740
SN will now form cluster by default as per ticket and Every n blocks ( where n can be configured through config var `block_interval` the leader node will transfer leadership to the top node retrieved by `masternodes top` API. Here is the logic
1- A goroutine spins up which keeps on checking after 30 seconds whether current node is a leader. (we can't just check 1 time & exit because a running node may become leader at some point.
2- Once it is a leader, it gets the current block count and enters the next "phase". 
3- In this Phase, it checks every 5 seconds the updated block count. If the difference b/w updated block count and block count from (2) equals or surpasses the config var `block_interval`, it then proceeds as in (4). Otherwise, it keeps on checking. Meanwhile, it also checks whether it is a leader or not.
4- It then calls masternodes top API, Gets the nodes and tries to get server id of each from the raft. Here, we loop through the results to find the first such node which has a valid server id.
5- With the server ID & Address obtained in (4), we call raft.TransferleadershipToServer() method. 

if transfer is unsuccesfull, it will keep on trying to transfer leadership. If not, it will go back to previous "phase" where it checks whether it has become leader or not every 30 seconds.

### DIFF
--- a/metadb/config.go
+++ b/metadb/config.go
@@ -28,6 +28,9 @@ type Config struct {
 
 	// data directory for rqlite
 	DataDir string `mapstructure:"data_dir" json:"data_dir,omitempty"`
+
+	// BlockInterval is the block interval after which leadership is transferred
+	BlockInterval int32 `mapstructure:"block_interval" json:"block_interval,omitempty"`
 }
 
 // SetWorkDir applies `workDir` to DataDir if it was not specified as an absolute path.
@@ -55,6 +58,9 @@ func (config *Config) Validate() error {
 	}
 	if config.DataDir == "" {
 		return fmt.Errorf("%s: %s", errValidationStr, "data_dir")
+	}
+	if config.BlockInterval == 0 {
+		config.BlockInterval = 48 // set default
 	}
 
 	return nil
--- a/metadb/config.go
+++ b/metadb/config.go
@@ -60,7 +60,7 @@ func (config *Config) Validate() error {
 		return fmt.Errorf("%s: %s", errValidationStr, "data_dir")
 	}
 	if config.BlockInterval == 0 {
-		config.BlockInterval = 48 // set default
+		return fmt.Errorf("%s: %s", errValidationStr, "block_interval")
 	}
 
 	return nil
@@ -68,5 +68,7 @@ func (config *Config) Validate() error {
 
 // NewConfig returns a new Config instance.
 func NewConfig() *Config {
-	return &Config{}
+	return &Config{
+		BlockInterval: 48,
+	}
 }
--- a/metadb/leader_transfer.go
+++ b/metadb/leader_transfer.go
@@ -1,0 +1,127 @@
+package metadb
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/pastelnetwork/gonode/common/errors"
+	"github.com/pastelnetwork/gonode/common/log"
+	"github.com/pastelnetwork/gonode/pastel"
+)
+
+// initLeadershipTransferTrigger is supposed to transfer leadership from current leader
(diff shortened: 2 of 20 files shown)


metadb/server.go
```diff
@@ -230,3 +239,24 @@ func (s *service) startServer(ctx context.Context) error {
...
+		return err
+	}
+
+	for _, nodeInfo := range nodeList {
+		address := nodeInfo.ExtAddress
+		segments := strings.Split(address, ":")
+		if len(segments) != 2 {
+			return errors.Errorf("malformed db node address: %s", address)
+		}
+		nodeAddress := fmt.Sprintf("%s:%d", segments[0], s.config.RaftPort)
```
mateeullahmalik, line 256: can you pls make sure we use raftport here and not http port @nguyen202 @danileng20.
related ticket: https://freedcamp.com/view/2956207/tasks/42099740

metadb/leader_transfer.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	for _, node := range nodes {
+		address := node.Address
+		segments := strings.Split(address, ":")
+		if len(segments) != 2 {
+			log.WithContext(ctx).WithField("address", address).Error("leadershipTransfer: malformed address")
+			continue
+		}
+		nodeAddress = fmt.Sprintf("%s:%d", segments[0], s.config.RaftPort)
```
mateeullahmalik, line 107: can you pls make sure we use raftport here and not http port @nguyen202 @danileng20 
related ticket: https://freedcamp.com/view/2956207/tasks/42099740

a-ok123 (dismissed):

supernode/examples/configs/mainnet.yml
```diff
@@ -25,6 +25,7 @@ metadb:
   http_port: 4001
   raft_port: 4002
   data_dir: "metadb"
+   block_interval: 48
```
nguyen202, line 28: nit : correct code indentation

metadb/config.go
```diff
@@ -56,6 +59,9 @@ func (config *Config) Validate() error {
 	if config.DataDir == "" {
 		return fmt.Errorf("%s: %s", errValidationStr, "data_dir")
 	}
+	if config.BlockInterval == 0 {
+		config.BlockInterval = 48 // set default
```
nguyen202, line 63: should set default config in `func NewConfig() *Config`.
Code here to validate, not set value.

metadb/leader_transfer.go
```diff
@@ -0,0 +1,134 @@
...
+		case <-time.After(leaderCheckInterval):
+			// this logic is only for leader, we don't want to exit for loop
+			// as current node can become leader in future
+			if !s.IsLeader() {
+				break
+			}
+
+			count, err := s.pastelClient.GetBlockCount(ctx)
+			if err != nil {
+				log.WithContext(ctx).WithError(err).Error("initLeaderElectionTrigger failed to call pastel client BlockCount")
```
nguyen202, line 33: in case of error, should break.

hoanlinh92 (approved):

metadb/leader_transfer.go
```diff
@@ -0,0 +1,135 @@
...
+				break
```
nguyen202, line 34: I think should retry again than break here?

a-ok123 (approved):

metadb/leader_transfer.go
mateeullahmalik: break only breaks switch case so it will try again after the interval (few seconds) ....so retrying here is not required

nguyen202: Thanks.

nguyen202 (approved):

PR #202, opened 2021-10-06
join cluster failure should not break sn

mateeullahmalik:
if join cluster fails, sn should not exit but it should retry to jin every 10 seconds

### DIFF
--- a/metadb/server.go
+++ b/metadb/server.go
@@ -16,20 +16,21 @@ import (
 )
 
 const (
-	defaultJoinAttempts            = 5
-	defaultJoinInterval            = 5 * time.Second
-	defaultRaftHeartbeatTimeout    = time.Second
-	defaultRaftElectionTimeout     = time.Second
-	defaultRaftApplyTimeout        = 10 * time.Second
-	defaultRaftOpenTimeout         = 60 * time.Second
-	defaultRaftWaitForLeader       = true
-	defaultRaftSnapThreshold       = 8192
-	defaultRaftSnapInterval        = 30 * time.Second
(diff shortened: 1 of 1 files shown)


metadb/server.go
```diff
@@ -143,30 +144,36 @@ func (s *service) initStore(ctx context.Context, raftTn *tcp.Layer) (*store.Stor
...
 }
 
+func (s *service) initClusterJoin(ctx context.Context, joinAddrs []string, interval time.Duration) {
+	log.WithContext(ctx).Infof("cluster-join initiated - joining in %v", interval)
+	raftAddr := fmt.Sprintf("%s:%d", s.config.ListenAddress, s.config.RaftPort)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
nguyen202, line 160: could you add a log here to know that this init code is existed due to ctx done?

metadb/server.go
```diff
@@ -143,30 +144,36 @@ func (s *service) initStore(ctx context.Context, raftTn *tcp.Layer) (*store.Stor
...
-			!s.config.NoneVoter,
-			defaultJoinAttempts,
-			defaultJoinInterval,
-			nil,
-		)
-		if err != nil {
-			return nil, errors.Errorf("join cluster at %v: %w", joinIPAddresses, err)
-		}
-		log.WithContext(ctx).Infof("successfully joined cluster at %v", joinAddr)
+		go s.initClusterJoin(ctx, joinIPAddresses, defaultJoinClusterRetryInterval)
```
nguyen202, line 147: Could you check again should allow metadb working if not yet joined a cluster?
In case required to join a cluster before start working, then should keep this code block here than in another go routine?

PR #204, opened 2021-10-09
fix self ip check

mateeullahmalik:


common/utils/utils.go
```diff
@@ -15,3 +20,21 @@ func SafeErrStr(err error) string {
...
+	}
+
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	return string(body), nil
```
nguyen202, line 39: it better to validate to ensure returned value is correct ip.

nguyen202 (approved):

common/utils/utils.go
mateeullahmalik: @nguyen202 its tried and tested method. how do you suggest to validate?

a-ok123 (approved):

PR #207, opened 2021-10-10
fix breaking sn

mateeullahmalik:


metadb/database/service.go
```diff
@@ -157,6 +156,15 @@ func (db *Ops) Run(ctx context.Context) error {
 	return nil
 }
 
+// Run run the rqlite database service
```
a-ok123, line 159: Nah, i think it should run again

metadb/metadb.go
```diff
@@ -53,5 +55,9 @@ func New(config *Config, nodeID string, pastelClient pastel.Client) MetaDB {
 
 // Run starts the rqlite server
 func (s *service) Run(ctx context.Context) error {
```
a-ok123, line 57: same - reset state, run again

PR #226, opened 2021-10-29
[p2p] wrong address of remote p2p peer

nguyen202:
Various fixes:
- Wrong receiver address
- Wrong sender address : should be externalIP than listen IP
- Fix type of `AppTicket     string    `json:"app_ticket"`` due to the change of base85
- Stop DHT when boostrap failed to retry
- Reply error code in P2P than stop without reply - taking time for request peer to wait for timeout
- Increase N - nearst IDs to 6 than 3
- Disable P2P secure encrytion - will enable later


pastel/reg_ticket.go
```diff
@@ -52,7 +52,7 @@ type NFTTicket struct {
 	Copies        int       `json:"copies"`
 	Royalty       float64   `json:"royalty"`
 	Green         bool      `json:"green"`
-	AppTicket     []byte    `json:"app_ticket"`
+	AppTicket     string    `json:"app_ticket"`
```
nguyen202, line 55: fix encode 85 issue.

p2p/kademlia/dht.go
```diff
@@ -227,8 +251,14 @@ func (s *DHT) Stats(ctx context.Context) (map[string]interface{}, error) {
 
 // new a message
 func (s *DHT) newMessage(messageType int, receiver *Node, data interface{}) *Message {
+	externalIP, _ := s.getExternalIP()
```
nguyen202, line 254: using external IP than listen IP

mateeullahmalik (approved):

a-ok123 (approved):

PR #235, opened 2021-11-12
[PSL-93] fix badger replication

mateeullahmalik:


a-ok123 (approved):

p2p/p2p.go
```diff
@@ -187,6 +188,8 @@ func (s *p2p) configure(ctx context.Context) error {
 	}
 	s.dht = dht
 
+	go s.store.InitCleanup(ctx, 5*time.Minute)
```
nguyen202, line 191: I don't think should put here.
Should start it just after complete initialization of store - like after https://github.com/pastelnetwork/gonode/blob/5d720253fa254b20580b3ab89739a157fc5525bb/p2p/p2p.go#L80.
To in case of ctx.Done() - then InitCleanup(ctx context.Context, cleanupInterval time.Duration)  routine also stops.
With your change - if any failed in following sections:
```
// start the node for kademlia network
	if err := s.dht.Start(ctx); err != nil {
		return errors.Errorf("start the kademlia network: %w", err)
	}

	if err := s.dht.ConfigureBootstrapNodes(ctx); err != nil {
		log.WithContext(ctx).WithError(err).Error("failed to get bootstap ip")

		return fmt.Errorf("unable to get p2p bootstrap ip: %s", err)
	}

	// join the kademlia network if bootstrap nodes is set
	if err := s.dht.Bootstrap(ctx); err != nil {
		// stop the node for kademlia network
		s.dht.Stop(ctx)
		// close the store of kademlia network
		s.store.Close(ctx)
		return errors.Errorf("bootstrap the node: %w", err)
	}

```
then p2p will run  configure(ctx context.Context) again - and create new store - leave un-managed InitCleanup() of previous store -> potential conflict.

mateeullahmalik: The client is thread safe so I don't believe we're going to have that problem. But I've still made the change. Thanks for pointing out. @nguyen202

nguyen202 (approved):

a-ok123 (approved):

PR #241, opened 2021-11-16
Integrate-storage-challenges

dnvu92:
The purpose of storage challenges is to able to verify a masternode is still save a symbol file hash or not. The basic flow as below:
1. challenger generate a challenge for every new block generated:
- Get merkle root in current block verbose 1
- Get list of rq symbol file keys from list of nft registration ticket
- Get challenging masternode id by calculate closest distance to merkleroot + random integer
- Get list of rq symbol file keys that contained in local node (this challenger)
- Get list of file hash to challenge by calculate closest distance to merkleroot + challenging masternode id
- Get responding masternode id (verifying masternode id) by calculate closest distance of masternode list that contains file hash to challenge and excluding challenging masternode id
- By each rq symbol file hash, generate random range of bytes data which is used to challenge (define by the start byte number and end byte number)
- By each symbol file hash, send the challenge message to challenging masternode
2. challenging masternode compute the hash of bytes range data from the local file and forward the challenge message result to the verifying masternode
3. verifying masternode check if the challenge message is expired or not (by checking the current block count and block count when challenge sent from challenger to challenging masternode), If not expired, the verifying masternode will validate the result hash of challenging masternode by using bytes range data of verifying masternode itself to hash and compare against with the responded hash from challenging masternode. If two hashes are the same, the challenging masternode passed the challenge, if not, the challenging masternode is dishonest.


nguyen202:
Could you provide link to spec + brief description of your change?
The changes at each module + purpose of each module? or  at least high level design?
It's hard to catchup this kind of big PR without any infomation.

messaging/protoactor_local.go
```diff
@@ -0,0 +1,58 @@
...
+
+// NewActor func
+func NewActor(system *actor.ActorSystem) Actor {
+	if lActor == nil {
+		lActor = &localActor{
+			context: system.Root,
+			mapPID:  make(map[string]*actor.PID),
+		}
+	}
+
```
nguyen202, line 56: this function is like singleton pattern.
I prefer change to follow example or like that:

func GetActorInstalnce(system *actor.ActorSystem) Actor {
	if lActor == nil {
		lActor = &localActor{
			context: actor.NewActorSystem().Root,
			mapPID:  make(map[string]*actor.PID),
		}
        return lActor
	}

messaging/protoactor_local.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Stop func
+func (r *localActor) Stop() {
+	for name, pid := range r.mapPID {
+		r.context.StopFuture(pid).Wait()
+		delete(r.mapPID, name)
+	}
+}
+
```
nguyen202, line 45: localActor's functions are not thread-safe, it's better to make it thread-safe

messaging/protoactor_remoter.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/AsynkronIT/protoactor-go/actor"
+	"github.com/AsynkronIT/protoactor-go/remote"
+	"github.com/gogo/protobuf/proto"
+	"github.com/pastelnetwork/gonode/common/context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+type remoteActor struct {
+	remoter *remote.Remote
```
nguyen202, line 13: remoteActor functions are not thread-safe, it's better to make it thread-safe

proto/supernode/storage_challenges.proto
```diff
@@ -0,0 +1,71 @@
...
+
+service StorageChallenge {
+  rpc GenerateStorageChallenges(GenerateStorageChallengesRequest) returns (GenerateStorageChallengesReply);
+  rpc ProcessStorageChallenge(ProcessStorageChallengeRequest) returns (ProcessStorageChallengeReply);
+  rpc VerifyStorageChallenge(VerifyStorageChallengeRequest) returns (VerifyStorageChallengeReply);
+}
+
+message StorageChallengeData {
+  string message_id = 1;
+  enum messageType {
```
nguyen202, line 18: why not define these enums outside of StorageChallengeData, and make it more compact, like:
```
enum MessageType {
    UNKNOWN = 0;
    STORAGE_CHALLENGE_ISSUANCE_MESSAGE = 1;
...
}
```
so we can have generated code:
```
type MessageType int32

const (
	MessageType_UNKNOWN                            MessageType = 0
	MessageType_STORAGE_CHALLENGE_ISSUANCE_MESSAGE MessageType = 1
)

```

supernode/services/storagechallenge/actor.go
```diff
@@ -0,0 +1,164 @@
...
+	return &domainActor{
+		conn: conn,
+	}
+}
+
+// OnSendVerifyStorageChallengeMessage handle event sending verity storage challenge message
+func (d *domainActor) OnSendVerifyStorageChallengeMessage(msg *sendVerifyStorageChallengeMsg) {
+	appCtx := msg.getAppContext()
+	for _, verifyingMasternodeAddr := range msg.VerifierMasternodesAddr {
+		log.Debug(verifyingMasternodeAddr)
```
nguyen202, line 92: could you add more context info to Debug message?
like log.WithField("verifyingMasternodeAddr", verifyingMasternodeAddr).Debug("OnSendVerifyStorageChallengeMessage")?

supernode/services/storagechallenge/generate_storage_challenge.go
```diff
@@ -0,0 +1,213 @@
...
+			ChallengeSliceCorrectHash:    "",
+			ChallengeResponseHash:        "",
+			ChallengeID:                  challengeID,
+		}
+
+		if err = s.sendprocessStorageChallenge(ctx, outgoingChallengeMessage); err != nil {
+			log.WithContext(ctx).WithError(err).Error("could not send process storage challenge")
+		}
+
+		s.repository.SaveChallengMessageState(ctx, "sent", challengeID, challengingMasternodeID, currentBlockCount)
```
nguyen202, line 143: in case of sending error, still SaveChallengMessageState()?

supernode/services/storagechallenge/generate_storage_challenge.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	return nil
+}
+
+func (s *service) sendprocessStorageChallenge(ctx context.Context, challengeMessage *ChallengeMessage) error {
+	masternodes, err := s.pclient.MasterNodesExtra(ctx)
+	if err != nil {
+		log.WithContext(ctx).WithField("challengeID", challengeMessage.ChallengeID).WithField("method", "sendprocessStorageChallenge").WithError(err).Warn("could not get masternode extra: ", err.Error())
```
nguyen202, line 152: already have .WithError(), so `: ", err.Error()` is redundant

common/utils/utils.go
```diff
@@ -73,6 +74,13 @@ func GetExternalIPAddress() (externalIP string, err error) {
 	return string(body), nil
 }
 
+// GetHashFromString generate sha256 hash from a given string
+func GetHashFromString(inputString string) string {
```
mateeullahmalik, line 78: there is already existing function for this in this file named `Sha3256hash(msg []byte) ([]byte, error)`

supernode/services/storagechallenge/verify_storage_challenge.go
```diff
@@ -0,0 +1,82 @@
...
+	} else if incomingChallengeMessage.ChallengeResponseHash == challengeCorrectHash {
+		challengeStatus = statusFailedTimeout
+		log.WithContext(ctx).WithField("method", "VerifyStorageChallenge").WithField("challengeID", incomingChallengeMessage.ChallengeID).Debug(fmt.Sprintf("masternode %s  correctly responded in %d blocks to a storage challenge for file %s, but was too slow so failed the challenge anyway!", incomingChallengeMessage.RespondingMasternodeID, blocksVerifyStorageChallengeInBlocks, incomingChallengeMessage.FileHashToChallenge))
+	} else {
+		challengeStatus = statusFailedIncorrectResponse
+		log.WithContext(ctx).WithField("method", "VerifyStorageChallenge").WithField("challengeID", incomingChallengeMessage.ChallengeID).Debug(fmt.Sprintf("masternode %s failed by incorrectly responding to a storage challenge for file %s", incomingChallengeMessage.RespondingMasternodeID, incomingChallengeMessage.FileHashToChallenge))
```
mateeullahmalik, line 44: use .WithFields() and pass a map

mateeullahmalik: fix at all places

mateeullahmalik (changes requested):
Left some comments & also agree with @nguyen202 feedback,

messaging/protoactor_local.go
dnvu92: Yes, you right.

PR #245, opened 2021-11-23
[PSL-116] change default reg confirmation

mateeullahmalik:


nguyen202:
I prefer change :
```
task.waitTxidValid(newCtx, task.regNFTTxid, int64(task.config.RegArtTxMinConfirmations)
```
to
```
task.waitTxidValid(newCtx, task.regNFTTxid, int64(task.config.RegArtTxMinConfirmations + 2)
```

How do you think?

mateeullahmalik:
any specific reason? I think we should use the variable

nguyen202:
> any specific reason? I think we should use the variable itself.
Because configuration is user's expectation. but actually should wait for more 2 confirmations to ensure all nodes are synced.

a-ok123 (approved):

nguyen202 (approved):

PR #258, opened 2021-12-08
[PSL-143] probe image reply

mateeullahmalik:


mateeullahmalik (commented):
@nguyen202 is out sick so I'm gonna go ahead and merged this as I need to build on it. 
Happy to make any changes later on if suggested

PR #259, opened 2021-12-09
[PSL-143] gen redundant IDs

mateeullahmalik:


pastel/dupe_detection.go
```diff
@@ -32,6 +32,10 @@ type DDAndFingerprints struct {
 	Maxes                      *DDMaxes               `json:"maxes"`
 	Percentile                 *DDPercentiles         `json:"percentile"`
 	Score                      *DDScores              `json:"score"`
+
+	// DD-Server does not directly return these comporessed fingerprints
+	// We generate and assign to this field to avoid repeated operations
+	ZstdCompressedFingerprint []byte `json:"zstd_compressed_fingerprint"`
```
nguyen202, line 38: should have `omitempty` in define it to prevent problem when un-marshall from json string

mateeullahmalik: added in next pr

nguyen202 (approved):

a-ok123 (approved):

walletnode/services/artworkregister/task.go
```diff
@@ -198,6 +213,62 @@ func (task *Task) run(ctx context.Context) error {
...
+	const digitBytes = "0123456789"
+	b := make([]byte, 4)
+	for i := range b {
+		b[i] = digitBytes[rand.Intn(len(digitBytes))]
+	}
+	randNumStr := string(b)
+
+	randNum, err := strconv.Atoi(randNumStr)
+	if err != nil {
+		return errors.Errorf("invalid random 4 bytes number: %w", err)
```
nguyen202, line 237: I think randUint32() is enough for 4 bytes with higher range.
And the error due to your random generator should not pause the registration.

pastel/dupe_detection.go
```diff
@@ -101,52 +105,65 @@ type PerceptualImageHashes struct {
...
 	if err != nil {
-		return "", errors.Errorf("GetProbeImageReply: failed to marshal dd-data: %w", err)
+		return nil, errors.Errorf("GetProbeImageReply: failed to marshal dd-data: %w", err)
 	}
 
-	resp = append(resp, separatorByte)
-	resp = append(resp, signature...)
+	res := make([]byte, base64.RawURLEncoding.EncodedLen(len(ddDataJson)))
+	encodedDdSig := make([]byte, base64.RawURLEncoding.EncodedLen(len(signature)))
+	base64.RawURLEncoding.Encode(res, ddDataJson)
```
nguyen202, line 157: pasteld is using stdEncoding, so we could change to base64.StdEncoding?

PR #266, opened 2021-12-15
Badger repalced with SQLite as storage for kademlia

a-ok123:


nguyen202 (approved):
LGTM!

p2p/kademlia/store/sqlite/sqlite.go
```diff
@@ -0,0 +1,201 @@
...
+
+	return nil
+}
+
+// Retrieve will return the local key/value if it exists
+func (s *Store) Retrieve(_ context.Context, key []byte) ([]byte, error) {
+	hkey := hex.EncodeToString(key)
+
+	r := Record{}
+	err := s.db.Get(&r, `SELECT data FROM data WHERE key LIKE ?`, hkey)
```
mateeullahmalik, line 117: shouldn't we use = here instead of LIKE?
LIKE is usually very expensive

p2p/kademlia/store/sqlite/sqlite.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, fmt.Errorf("failed to get record by key %s: %w", hkey, err)
+	}
+	return r.Data, nil
+}
+
+// Delete a key/value pair from the Store
+func (s *Store) Delete(ctx context.Context, key []byte) {
+	hkey := hex.EncodeToString(key)
+
+	res, err := s.db.Exec("DELETE FROM data WHERE key LIKE ?", hkey)
```
mateeullahmalik, line 128: same

p2p/kademlia/store/sqlite/sqlite.go
a-ok123: `=` didn't work at all

mateeullahmalik (approved):

p2p/kademlia/store/sqlite/sqlite.go
a-ok123: although, it didn't work when i was writing key as is, now i convert it to hex string, i will re-test

mateeullahmalik: ok - I've approved just in case

nguyen202 (approved):

mateeullahmalik (approved):

PR #276, opened 2021-12-21
[NOJIRA] add json rpc id, provent reuse request

dnvu92:


nguyen202:
Sorry, not sure why we need to have ID per request? Could you put the reason of any change in description?

dnvu92:
> Sorry, not sure why we need to have ID per request? Could you put the reason of any change in description?
The json rpc id used to match responses to requests when using asynchronous or batch calls.
It's fine whenever call single request at a time, but sometimes, the system call many request in the same time. So the `id` is needed in this case

nguyen202:
> > Sorry, not sure why we need to have ID per request? Could you put the reason of any change in description?
> 
> The json rpc id used to match responses to requests.
> 
> It's fine whenever call single request at a time, but sometimes, the system call many request in the same time.

pasteld supported it? or for debugging?

mateeullahmalik (approved):

nguyen202 (approved):

PR #278, opened 2021-12-26
[PSL-176]prepare sense code base

nguyen202:
https://pastel-network.atlassian.net/browse/PSL-176
- Clone flow from registration flow
- Removed steps related to thumbnail conversion, raptorq symbols calculation + storage
- Update proto file + re-generate  to put output golang files into different sub packages instead of same packages


bacnh8 (approved):
LGTM - I will add again the flows of step 3 again into this flow at other PR.

mateeullahmalik (approved):

PR #280, opened 2021-12-27
[PSL-183] kademila regression testing

mateeullahmalik:


a-ok123 (dismissed):

integration/configs/sn1.yml
```diff
@@ -0,0 +1,47 @@
...
+
+p2p:
+  listen_address: "0.0.0.0"
+  port: 14445
+  data_dir: "p2p-data"
+  bootstrap_ips: "snserver-2:14445,snserver-3:14445"
+  external_ip: "snserver-1"
+
+metadb:
+  is_leader: false
```
nguyen202, line 32: should set it to `true`

integration/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+  restart: "${COMPOSE_RESTART_BEHAVIOR:-on-failure}"
+x-network: &network
+  networks:
+    - internal
+x-default: &default
+  <<: *restart_policy
+  <<: *network
+
+services:
+  sn1:
```
nguyen202, line 13: it better to have SN1 started finish first, delay some seconds before starting SN2/SN3 to SN1 to be a metadb leader

integration/integration_suite_test.go
```diff
@@ -0,0 +1,66 @@
+package integration_test
+
```
nguyen202, line 2: could you organize `integration` folder into sub folders like:
```
./
├── utility
│   └── helper.go
├── test
│   ├── p2p
│   ├── metadb
│   ├── ...
 ```

Comment (mateeullahmalik) [reply integration/integration_suite_test.go:2]: ok - I'll do in next PR

File: integration/configs/sn1.yml (hunk 1 above)
Comment (mateeullahmalik) [reply integration/configs/sn1.yml:32]: we are not using meta db - these tests only work with p2p atm

File: integration/docker-compose.yml (hunk 2 above)
Comment (mateeullahmalik) [reply integration/docker-compose.yml:13]: we are not worried about metadb right now

File: .circleci/config.yml [hunk 4]
```diff
@@ -1,3 +1,5 @@
+orbs:
+  go: circleci/go@1.5.0
```
nguyen202, line 2: could you check again version of go? we're using 1.7

mateeullahmalik: yes - but I had to keep 1.5.0 for go circle ci executor as some libraries versions were mismatching. The circleci orbs for 1.7.0 doesn't seem to be working fine right now. 
Besides, only the testing code is gonna run with this executor and not our actual code. 
I'll keep an eye on circleci go executor updates and update in when it makes sense

a-ok123 (approved):

Dicklesworthstone (approved):

PR #283, opened 2021-12-30
WIP:[PSL-195] Storage challenge integration test

dnvu92:


mateeullahmalik (changes requested):
Please take a look at gonode/integration
I want us to follow that & build upon that instead of defining more  (possibly duplicate) integration testing in different parts of the code base. 
I don't mean to say what you did is wrong - I definitely think some part of this can can used.

PR #308, opened 2022-02-25
[PSL-195] Storage challenge refactored

Bjbrostoff:
The goal of the storage challenge system is to ensure that supernodes are hosting the files they should be.  In order to do this, storage challenges are sent out with each new block.  For each storage challenge, multiple challengers are deterministically selected, who then deterministically select: which supernodes to check, which files to check, and which other supernodes should verify those checks.

The current state of the storage challenge system:
Storage challenge is functioning, but if a storage challenge verification fails, nothing happens.
Callbacks in task/stateStorage can be adjusted when proper consequences are determined.


Bjbrostoff:
Closing this pull request as code was integrated with another branch.

PR #316, opened 2022-04-04
[PSL-247] increase connection deadline & pool size

mateeullahmalik:


a-ok123 (approved):

Bjbrostoff (approved):
Lools like not much has changed - that connection deadline being 60 minutes may remove its intended functionality.  If it's what is needed right now to get the system working, I will approve.

PR #321, opened 2022-04-11
[PSL-314] file size round up for action fee calculation

mateeullahmalik:


Bjbrostoff (approved):
Approved pending integration tests

PR #329, opened 2022-04-18
Psl 316 expand status tracking

Bjbrostoff:
- Added the ability to get the history of a task via REST API
- Added the ability to create "ephemeral" statuses within a task, which will post to the same status channel.  With this, we can get transaction ID's for sense and cascade as the task moves along.
- Implemented getting transaction ID's for sense and cascade as the task proceeds.


a-ok123:
Also, we need to add the same repotting for NFT registration

mateeullahmalik (approved):

PR #369, opened 2022-10-13
[PSL-414] implement db and api structural changes to support log details

j-rafique:


common/service/task/state/state.go
```diff
@@ -141,3 +153,13 @@ func New(subStatus SubStatus, taskID string) State {
 		store:  store,
 	}
 }
+
+func isValid(statusLog map[string]interface{}) bool {
+	if statusLog != nil {
+		if len(statusLog) != 0 {
```
mateeullahmalik, line 159: I Think isValid should only check nil map and just to check nil map.. do we really need a function? 
I think we can improve the logic around here
I would also think about  defining our own type like type Fields map[string]interface{} and then writing function for it. 

func (f Fields) IsValid() bool

if not, then please consider `return len(statusLog) != 0`

common/storage/local/sqlite.go
```diff
@@ -56,14 +66,21 @@ func (s *SQLiteStore) QueryTaskHistory(taskID string) (history []types.TaskHisto
...
-		err = rows.Scan(&i.ID, &i.CreatedAt, &i.TaskID, &i.Status)
+		var details string
+		err = rows.Scan(&i.ID, &i.CreatedAt, &i.TaskID, &i.Status, &details)
 		if err != nil {
 			return nil, err
 		}
+
+		err = json.Unmarshal([]byte(details), &i.Details)
+		if err != nil {
+			fmt.Errorf("cannot unmarshal task history details: %w", err)
```
mateeullahmalik, line 77: you mean log.Error()? or return err
I won't return err though.. just log

mateeullahmalik (approved):

PR #370, opened 2022-10-14
[PSL-412] use more nodes for downloading valid files sense, cascade &…

mateeullahmalik:
… nft


walletnode/services/download/task.go
```diff
@@ -143,14 +144,28 @@ func (task *NftDownloadingTask) Download(ctx context.Context, txid, timestamp, s
...
+			}
+		}
+
+		if matches >= 2 {
+			return a, nil
 		}
 	}
-	return nil
+
+	return 0, errors.Errorf("no file matched other two files, no. of files - %d", len(task.files))
```
a-ok123, line 168: maybe better message - "Can't find three matching files" or smth like that?

a-ok123 (approved):

a-ok123 (approved):

PR #539, opened 2023-05-09
Update fingerprints.go

Dicklesworthstone:


mateeullahmalik:
I don't think we need these changes as we already have exactly same structure. Please correct me if I'm wrong

